PR #723, opened 2026-06-02
Design: Agent Cognition Core (memory + rules engine + tools layer)

brandonkindred:
## What

Adds a **Feature Specification** for the **Agent Cognition Core** — a reusable substrate that
upgrades agents generated by the Agentic AI team from bare LLM wrappers into agents with the
faculties of a reasonable thinking person:

- **Memory layer** — time-partitioned episodic events with idempotent day → week → month →
  year rollups and budget-aware retrieval.
- **Rules engine** — hybrid `advisory` (prompt-injected) + `enforced` (deterministic
  pre/postcondition) rules, learned from memory and **proposed for human approval** before
  taking effect (HITL).
- **Tools layer** — per-agent tool binding composed from existing tool/integration registries
  and the shared tool loop; every tool call feeds memory.

## Files

- `backend/agents/agent_cognition/DESIGN.md` — full spec with architecture, use-case,
  logic-flow (invoke path + rollup/derivation), rule-lifecycle, and security diagrams (Mermaid).
- `backend/agents/agent_cognition/README.md` — package overview pointing at the spec.

## Key design decisions

- Cognitive state lives in the long-lived **platform Postgres** (namespaced by `agent_id`),
  never in the ephemeral per-agent sandbox (`docker compose down -v` deletes its volumes).
- **Inject-on-invoke / writeback-on-return**: the proxy injects rules + memory digest into the
  invoke payload and persists the returned writeback, so sandboxed/network-isolated agents
  never need direct store access.
- Rollups run via a **central scheduler + lazy catch-up** on invoke (shared idempotent,
  upsert-keyed function).
- Heavy reuse of existing infra (`shared_postgres`, `llm_service`, `agent_console.prune`
  pattern, `agent_llm_tools_service`, `tool_loop`); the only net-new pieces are the rules
  engine and the cognition orchestration seam.

## Scope

**Design only** — no runtime code in this PR. Building the modules, wiring the core into the
generator/`AGENT_ANATOMY.md`, and the HITL review UI are tracked as follow-ups in the spec's
"Open items / out of scope" section.

> Note: no linked tracker issue exists for this design effort yet, so no `Closes #N` is
> included. Happy to open one and link it if you'd like.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TYV6X5v9Rcrdu5pyDzrZxa)_


brandonkindred (commented):
Addressed both findings in `ffb4549`.

**Gateway coverage (r3344309131):** confirmed — `_is_team_path()` matches only `TEAM_CONFIGS` prefixes via `_get_team_prefixes()`, so `/api/agents/...` is not gated. The operator routes now move to a dedicated **`/api/cognition/...`** prefix that must be added to the gateway's matched-prefix set, with an explicit §10 note and a "Gateway alignment" control update. Step 12's acceptance now requires a gateway test proving the cognition prefix is intercepted.

**Postcondition before persist (r3344309138):** confirmed — the §8.1 sequence persisted writeback before the check (and disagreed with the §11 flow). Reordered so enforced postconditions run **before** `persist_writeback`; on violation only a sanitized blocked-run audit event is persisted, never the rejected `cognition_writeback`. Added a callout in §8.1, a "Writeback integrity" control in §11, and updated Step 10's acceptance.

---

brandonkindred:

brandonkindred (commented):
All six findings from this round addressed in `e7233b4` (design + plan). Each was valid.

**P1 — Operator authorization (r3344493301):** `resolve_author` is now explicitly labeled *provenance only* (can return `anonymous`, never an authz decision). `approve`/`reject` require an explicit operator-authorization check — `COGNITION_OPERATOR_TOKEN` (constant-time compare) now, operator role when platform auth lands — added as a §11 control, the §10 endpoint note, Step 12 acceptance (401/403 without credential), and the Step 13 env list.

**P1 — Platform-mediated tool path for sandboxes (r3344493305):** added a "Tool execution topology" subsection. Tools resolve to an execution site at bind time: `in_process`, `sandbox_local` (run inside the sandbox; v1 default), or `platform_bound`. Platform-bound tools used by sandboxed agents run via a **proxy-driven multi-turn SB↔PX `tool_calls`/`tool_results` loop** so secrets/egress stay platform-side — wired into Step 7 + Step 10.

**P2 — Proposal actions for retire/amend (r3344493314):** `agent_cognition_rule_proposals` now carries `action (add|retire|amend)` + `target_rule_id`. Approve applies the action deterministically (add→insert active; retire→retire target; amend→retire target + insert replacement). Documented in §8.3 and Steps 5/6.

**P2 — Recompute on late events (r3344493319):** summaries gain `covers_through` + `stale`. A late event whose `occurred_at` falls in an already-summarized period sets that period `stale` and cascades up the parent scales; the rollup processes "not-yet-summarized **or** stale" periods and recomputes. §8.2 note + Steps 2/3.

**P2 — Digest vs closed-period rollups (r3344493321):** the digest now uses the latest **closed** day/week/month summaries plus the **in-progress** period represented directly from recent raw events — so mid-period invokes are never context-starved. Partial rollups noted as optional/out-of-v1. Step 4 updated.

**P2 — Idempotent writeback across retries (r3344493322):** events are keyed on `(agent_id, source_run_id, source_seq)` (`source_run_id` = invoke `trace_id`) with `ON CONFLICT DO NOTHING`, so a retried/re-delivered writeback is a no-op. §8.1 note, §9 schema, §11 control, Step 2.

---

brandonkindred:

brandonkindred (commented):
All seven findings from this round addressed in `0ec4ce9` (design + plan). Each was valid.

**P1 — Stable run id for idempotency (r3344544643):** correct — the shim mints `trace_id` per POST, so a retry wouldn't conflict. `source_run_id` is now defined as a **proxy-minted stable idempotency key for the logical invoke** (reusing a caller `Idempotency-Key` when present), propagated into the sandbox and reused across retries — explicitly *not* the shim's `trace_id`. Updated §8.1, §9, §11, Step 2, Step 10.

**P1 — Don't roll months up from weeks (r3344544647):** correct — ISO weeks straddle months. Rollup inputs are now calendar-scoped: day←events, week←days, **month←days**, year←months. Added a "Rollup input scoping" note (§8.2) and updated the §8.2 diagram and Step 3.

**P2 — Events unique constraint in the schema step (r3344544649):** Step 1 now explicitly declares the **events unique `(agent_id, source_run_id, source_seq)`** constraint (plus summaries `version`/`stale` and evidence columns) so Step 2's `ON CONFLICT` target exists; acceptance updated.

**P2 — Keep tool-call audit on blocked outputs (r3344544651):** correct — tools that already ran are real side effects. The postcondition-blocked path now persists **sanitized `tool_call`/`outcome` records** (preserving "every tool call feeds memory") while dropping the rejected output and untrusted memory. Updated §8.1 diagram + note, the §11 flow (new `BLKO` node) and control, and Step 10.

**P2 — Operator auth for reads too (r3344544652):** the operator-authorization requirement now covers **all** cognition routes — the `GET` reads as well as `approve`/`reject` — since memory/rules/proposal evidence are private state. Updated §10 note, §11 control, Step 12 acceptance.

**P2 — Invalidate stale-evidence proposals/rules (r3344544656):** summaries now carry a `version` bumped on recompute; proposals and derived rules store `(summary_id, version)` evidence refs. When a referenced summary's version advances, pending proposals are flagged `stale_evidence` (re-eval/withdraw) and active derived rules get `needs_review` (resurfaced to the operator). Added a `reevaluate_stale` hook. Updated §8.2 note, §9, §11, Steps 3 & 6.

**P3 — Use the modeled `rejected` status (r3344544659):** Step 5 now lands `reject → rejected` (matching the `pending|approved|rejected` model, API filters, and the §8.3 lifecycle) instead of an undocumented `archived`.

---

brandonkindred:

brandonkindred (commented):
Fixed in `4d50d1c`. The `reject` endpoint's Purpose column now reads "reject proposal (status → `rejected`)" and UC6 reads "rejects → proposal `rejected`", so the §10 API table and §6 use-case match the data model (`pending|approved|rejected`), the §8.3 lifecycle, and the implementation plan — status filters (`status=pending`/`rejected`) won't diverge from stored state. The only remaining "archived" reference is the plan's explicit "not an undocumented `archived`" guidance, which is intentional.

---

brandonkindred:

brandonkindred (commented):
All three findings from this round addressed in `0d25df0` (design + plan). Each was valid.

**P1 — Idempotent retries re-running side effects (r3344729122):** correct — deduping only `persist_writeback` still let the sandbox (and its tool calls) run twice. Idempotency now gates **execution**: a new `agent_cognition_runs` ledger (PK `(agent_id, source_run_id)`) is claimed before invoking; a retry with a `completed`/`blocked` key **replays the stored envelope without re-invoking the sandbox**, and a retry arriving while `in_progress` gets `409`. The writeback unique key remains as defense-in-depth, and the unavoidable lost-ack-before-commit window is explicitly covered by requiring platform-bound tool handlers to be idempotent where feasible (anything without it is documented at-least-once). Updated §8.1 (note + sequence), §9 (`agent_cognition_runs`), §11, Steps 1/8/10.

**P2 — Cognition context polluting the user input contract (r3344729127):** correct — `dispatch.invoke_entrypoint` passes the body straight to the entrypoint, so a sibling `cognition` key would break strict/non-object request models. The proxy now sends a **wrapper envelope `{input, cognition}`** (output mirrors `{output, cognition_writeback}`); `mount_invoke_shim`/`dispatch` unwrap it and invoke the entrypoint with `input` **only**, exposing `cognition` via a side channel, with passthrough for non-enveloped bodies. This is an explicit `shared_agent_invoke` change in Step 7. Updated §10, §11 (Input-contract isolation), §7 diagram, Steps 7/8/10.

**P2 — Terminal state for withdrawn stale proposals (r3344729131):** correct — `reevaluate_stale` had no valid terminal status. Added a **`superseded`** status (`pending|approved|rejected|superseded`): the system auto-withdraw state for stale-evidence proposals, distinct from `rejected` (a human decision) so it neither fabricates an operator rejection nor lingers in the queue. Updated §9, §8.2/§8.3 (lifecycle diagram gains `pending → superseded`), Steps 1/6.

Both revised diagrams (invoke sequence with the run-ledger branch; rule-lifecycle with `superseded`) validate.

---

brandonkindred:

brandonkindred (commented):
All seven findings from this round addressed in `0a4cded` (design + plan). Each was valid; several caught real architectural gaps.

**P1 — Lease for in-progress claims (r3344833328):** correct — a crashed proxy would wedge the key at `409` forever. The `agent_cognition_runs` row now carries `lease_expires_at`; an `in_progress` claim past its lease is reclaimable (a retry re-executes), and the scheduler also reclaims expired rows. Documented as at-least-once for the reclaim window (tool idempotency covers it).

**P2 — Compare retry payloads before replaying (r3344833338):** the ledger now stores a `request_hash`; a retry whose body differs from the original → `409` (never serve the prior response for a different request); only a matching hash replays.

**P2 — Disambiguate envelopes from user inputs (r3344833344):** correct — keying off the presence of `input` was unsafe. The envelope now uses an explicit namespaced marker `__khala_cognition_envelope__`; the shim unwraps **only** when the marker is present and well-formed, and passes any unmarked body (even one with its own `input` key) through unchanged. The proxy rejects a caller body that already contains the marker.

**P2 — Enforce configured event retention (r3344833358):** correct — `retention_days_events` had no teeth. Added a `prune_events` store method and a scheduler pruning pass that deletes raw events past the cutoff **only** once folded into a non-stale day summary (summaries retained). §8.2 + Steps 2/11.

**P2 — Trusted audit source for sandbox-local tools (r3344833362):** correct — for sandbox-local tools nothing crossed the boundary, so the blocked-run "persist tool audit" promise had no trusted source. The shim now **brokers/wraps** sandbox-local tool handlers and emits a trusted out-of-band tool-call audit; blocked runs persist the **shim** audit, not the model's writeback. Tools invoked without the broker are explicitly out-of-scope.

**P2 — Inject advisory rules where prompts are built (r3344833366):** correct — the proxy has no shared prompt seam (`dispatch` calls an arbitrary entrypoint). Reframed: the proxy enforces only the deterministic enforced rules; advisory rules ride the `cognition` side channel and the **generated agent runtime renders them into each LLM call** (now a Step 14 generator responsibility). Advisory rules are explicitly best-effort; only enforced rules are guaranteed.

**P2 — Recheck payload limits after adding cognition (r3344833371):** `AGENT_INVOKE_MAX_PAYLOAD_BYTES` is now applied to the assembled `{marker, input, cognition}` envelope (and the digest is token-budgeted), so attaching cognition can't push a near-limit request over the cap.

While revising I also caught and fixed a latent Mermaid bug: semicolons in sequence-diagram message text were being parsed as statement separators (the §8.2 diagram had been invalid since an earlier round) — replaced with commas/words; all diagrams now validate.

---

brandonkindred:

brandonkindred (commented):
All three findings from this round addressed in `d270a7f`. All valid — and notably two of them caught bugs that my own round-6 fixes introduced (retention vs. recompute, and lease reclaim dropping the hash). Good catches.

**P1 — Preserve rollup inputs before pruning raw events (r3344917207):** correct — the new retention pruner collided with the late-arrival recompute: a backdated event in an already-pruned period would have rebuilt the summary from only the late row. Fixed with **two regimes** (§8.2): a late event in a **retained** period → recompute from events (as before); a late event in an **already-pruned** period → **incremental amend** (`revise_summary(base_summary, [late_events])` extends the existing summary, never rebuilds from scratch), so pruned history is preserved. Pruning is still only allowed once the day summary exists and is non-stale, so an amend always has a base. Updated §8.2, §11, Steps 2/3.

**P1 — Preserve expired ledger rows for idempotency checks (r3344917211):** correct — having the scheduler delete expired-lease rows dropped the only stored `request_hash`, so a post-expiry retry with a different body couldn't be told from first sight. Fixed: the scheduler now **GCs only terminal (`completed`/`blocked`) rows past an idempotency TTL and never touches `in_progress` rows**; expired-lease reclaim is done **lazily and in place by `claim_run`**, which resets the lease but **retains the original `request_hash`** and still compares it (different body → 409, same body → re-execute). Updated §8.1, §8.2, §11, Steps 8/11.

**P2 — Make generator wiring depend on the invoke proxy (r3344917212):** correct — Step 14's acceptance needs the cognition side channel, which only exists once Step 10 wraps/unwraps the envelope. Step 14 now **depends on 9, 10, 13**, and the dependency graph has the `S10 → S14` edge.

Both revised diagrams (rollup/retention sequence, dependency graph) validate.

---

brandonkindred:

brandonkindred (commented):
All six findings from this round addressed in `fdab72f`. Each was valid.

**P1 — Enforce forbidden tools before execution (r3344998871):** correct — a `forbid_tool` checked only as an output postcondition runs too late. The shim tool-broker now evaluates the active enforced tool-restriction predicates (`evaluate_tool_call`) **before dispatching each handler**, refusing a disallowed call pre-dispatch (recorded as a blocked tool-audit event). Output postconditions stay for after-the-fact value/shape checks. §8.1 (note + `SB->>SB` gate), §7 topology, §11, Steps 5/7.

**P1 — Store blocked envelopes for idempotent replay (r3344998874):** correct — the replay contract covers `blocked`, but the blocked path stored no envelope. Both block paths now **persist the 4xx response envelope** into the ledger, so a retried precondition/postcondition block replays the same 4xx. §8.1, §9, §11, Step 10.

**P2 — Require caller keys for run-once side effects (r3344998882):** correct — minting a fresh id when the caller omits `Idempotency-Key` makes retries look new. Now: with no caller key the proxy keys on `request_hash` (byte-identical retries still dedup), and a side-effecting agent can set the new manifest flag `requires_idempotency_key: true` to **require** a caller key (else `400`). Without a key and without the flag, the call is explicitly **at-least-once**, not run-once. §8.1, §10 manifest, §11, Steps 9/10.

**P2 — Add the runtime turn protocol before platform tools (r3344998890):** correct — the proxy-driven loop needs a runtime that pauses/emits tool calls/resumes, which the single-shot `dispatch.invoke_entrypoint` lacks. Documented the **multi-turn runtime protocol** as a prerequisite: Steps 7/10 land it with a stubbed-runtime test; live `platform_bound` tools for *generated* agents are gated on the Step 14 runtime scaffold; **`sandbox_local` is the v1 default** (needs no protocol). §7, Step 7.

**P2 — Keep `superseded` in the proposal status model (r3344998895):** Step 5's acceptance text still said `pending|approved|rejected`. Corrected everywhere to **`pending|approved|rejected|superseded`** as a valid, filterable status. Step 5.

**P2 — Cap user output separately from writeback (r3344998898):** correct — sharing one `AGENT_INVOKE_MAX_OUTPUT_BYTES` over `{output, cognition_writeback}` lets control data starve user output (or drop memory). The cap is now **per-field**: `AGENT_INVOKE_MAX_OUTPUT_BYTES` on `output`, a separate `AGENT_COGNITION_WRITEBACK_MAX_BYTES` on the writeback (truncate + `truncated: true` flag). §10, §11, Steps 10/13.

Both revised diagrams (invoke sequence with the pre-dispatch tool gate; dependency graph with `S5 → S7`) validate.

---

PR #821, opened 2026-06-08
Extract BarPredicate IR into a dedicated module with snapshot tests

### LINKED ISSUE #471: indicator_probe: introduce a BarPredicate IR layer between extraction and aggregation (#406)
## Context

PR #456 ships the indicator-coverage probe (#448) for the #406 epic. The implicit contract between `_extract_subconditions` (the AST walker) and `_aggregate` (the report builder) is currently encoded as **boolean flags on `_Group`**:

```python
class _Group:
    subconds: list[_Subcond]
    target_symbols: Optional[set]
    combinator: Literal["and", "or"]
    ancestor_count: int
    has_unknown_or_leg: bool
    has_unknown_and_conjunct: bool
```

…plus a `has_unknown_leg: bool` on individual `_Subcond`s. This flag soup encodes structural information (AND vs OR, where ancestors end and OR alternatives begin, which legs are unknown) that an explicit IR would express directly via tree shape.

This is option #7 of the complexity-reduction plan attached to PR #456. **It is the largest design change and should land as a separate PR after #COMPOUND-VISITOR + #COVERAGE-AGGREGATOR.**

## Problem

Every recent review cycle on PR #456 has surfaced bugs in flag-handling — propagating `has_unknown_and_conjunct` into nested bodies (#3213815817), interpreting `has_unknown_or_leg` correctly across symbol gates (#3213748113), preserving symbol filters across OR/AND boundaries (#3197890414), and so on. The flags are correct individually but their interaction surface is large because they're a denormalised projection of an underlying tree.

A new combinator (XOR, NOT, implication) currently means a new flag and updates to every aggregator branch that reads the existing flags.

## Proposed approach

Define an explicit IR (Python 3.10 `match`/`case`):

```python
@dataclass(frozen=True)
class AndOp:
    legs: tuple[BarPredicate, ...]
    unknown: bool

@dataclass(frozen=True)
class OrOp:
    legs: tuple[BarPredicate, ...]
    unknown: bool

@dataclass(frozen=True)
class SymbolGate:
    syms: frozenset[str]
    inner: BarPredicate

@dataclass(frozen=True)
class MaskLeaf:
    label: str
    evaluator: Callable[[pd.DataFrame], pd.Series]

@dataclass(frozen=True)
class Static:
    value: bool  # True = no-op, False = unreachable

BarPredicate = AndOp | OrOp | SymbolGate | MaskLeaf | Static
```

- Extractor produces a tree of these.
- Aggregator pattern-matches:
  ```python
  def evaluate(pred: BarPredicate, df: pd.DataFrame) -> pd.Series:
      match pred:
          case MaskLeaf(_, fn): return fn(df)
          case AndOp(legs, _):  return reduce(operator.and_, (evaluate(l, df) for l in legs))
          case OrOp(legs, _):   return reduce(operator.or_,  (evaluate(l, df) for l in legs))
          ...
  ```
- The `_Group.combinator` / `ancestor_count` / `has_unknown_or_leg` / `has_unknown_and_conjunct` flags collapse into tree shape.

## Expected impact

- ~200 LoC trim, mostly from eliminated flag-handling branches in `_aggregate`.
- Largest future-extensibility win: new combinators = new IR variants, not new flags + audits of every reader.
- Extractor and aggregator become independently fuzzable (the IR is a stable middleman).
- `coverage_probe/runtime_instrument.py` and any future probe variant could share the IR.

## Risk

**Highest design risk among the refactor options.** An IR is a new internal contract whose bugs may not surface at the report level. Mitigation:

- **Snapshot-test the IR for ~20 representative strategies** before deleting `_Group` flags. Assert byte-identical IR pre/post the refactor by running both the new tree-builder and the old flag-emitter on the same input.
- Land in two stages:
  1. Introduce the IR alongside the existing flags; assert equivalence at every emission site.
  2. After a soak period, delete the flags.

## Acceptance

- [ ] `BarPredicate` IR module added (`coverage_probe/predicate_ir.py`).
- [ ] Snapshot tests for 20 representative strategies asserting IR ↔ flag-encoding equivalence.
- [ ] `_aggregate` (post-#COVERAGE-AGGREGATOR) pattern-matches the IR; `_Group` flags removed.
- [ ] All 198 indicator-probe + static-probe tests pass.
- [ ] `ruff check` + `ruff format --check` clean.

## Order

Lands strictly after:
- The `SubconditionVisitor` refactor (visitor produces the IR cleanly).
- The `CoverageAggregator` refactor (aggregator already class-shaped, easier to migrate to pattern matching).

Should be its own PR — do not bundle with options 1–6.

## References

- Parent epic: #406
- Parent issue: #448
- Originating PR: #456
- Plan file: `/root/.claude/plans/there-are-over-9000-drifting-dahl.md`

brandonkindred:
Closes #471

## Context

The `BarPredicate` tree IR that replaces the old `_Group`/`_Subcond` flag soup **already shipped** with the prerequisite `SubconditionVisitor` (compound-visitor) and `CoverageAggregator` refactors. In the current code the IR nodes, the `PredicateGroup` container, and the pure tree-operation helpers all live, and `_aggregate` already walks the tree — the `combinator` / `ancestor_count` / `has_unknown_or_leg` / `has_unknown_and_conjunct` / `has_unknown_leg` flags are gone.

So three of the five acceptance items were already satisfied. This PR completes the remaining two:

1. The IR still lived **inline** in `indicator_probe.py`; the issue calls for a dedicated `coverage_probe/predicate_ir.py` module so the IR is a standalone, independently testable contract.
2. There were no snapshot tests pinning the IR shape. The original "IR ↔ flag-encoding equivalence" framing is moot (the flag encoding no longer exists to compare against), so it's reframed as snapshotting a **canonical render of the extracted IR tree** for ~20 representative strategies — a direct regression guard on the extractor.

## What changed

- **New `coverage_probe/predicate_ir.py`** — moves the IR nodes (`MaskLeaf`, `Static`, `SymbolGate`, `AndOp`, `OrOp`, `Leg`, the `BarPredicate` union) and the `PredicateGroup` container out of `indicator_probe.py`, alongside the pure tree utilities. Because those helpers form the cross-module contract between the extractor and the IR layer, they are exported as **public API** (no leading underscore): `eval_tree`, `collect_legs`, `tree_and_unknown` / `tree_or_unknown`, `tree_effective_symbols`, `find_or_groups`, `leg_gate_symbols`, and `build_and_group` / `build_or_group`. Only the `*_walk` recursion helpers stay private to the module. The module depends only on `pandas` + the stdlib — no `ast`, no coverage-report models — so the extractor and aggregator can be exercised independently through it. Code moved with no logic changes; `indicator_probe.py` imports the symbols back.
- **Canonical renderer** — `render_bar_predicate` / `render_predicate_group` emit a deterministic, identity-independent text form of the IR (leaves render by label only since the evaluator is a closure; symbol sets are sorted). `render_predicate_group` uses an `is not None` check so an explicit (even empty) denylist is never conflated with `None`.
- **New snapshot suite** `test_predicate_ir_snapshot.py` — pins the extracted IR render for 20 representative strategies (golden in `tests/golden/snapshots/predicate_ir.json`, regenerated **only** with `UPDATE_PREDICATE_IR_SNAPSHOTS=1`; a missing golden without the flag fails loudly rather than being silently regenerated). Live-render extraction and golden loading run in session-scoped fixtures, so the work happens at test time rather than at import/collection. Includes focused unit tests for every public IR helper and the renderer branches.

## Review follow-ups

- Promoted the cross-module tree helpers to public names (dropped the leading underscores) so `indicator_probe.py` no longer imports module-private symbols; the module docstring now lists the public surface.
- Removed the unused `_strip_symbol_gates` helper — it had no callers tree-wide — along with its unit test, rather than carrying it as dead code. (It is **not** present in `predicate_ir.py`.)
- Gated snapshot regeneration strictly behind `UPDATE_PREDICATE_IR_SNAPSHOTS`, computed the live renders once, switched the `denied_symbols` render to `is not None`, and documented `MaskLeaf` hashability.

## Verification

- `ruff check` + `ruff format --check` clean on the `coverage_probe/` package and the IR tests.
- Investment Strategy Lab CI job green; all indicator-probe + coverage-probe + static-probe + IR snapshot suites pass.
- `predicate_ir.py` at **100%** line coverage; the `coverage_probe` package stays comfortably above the 90% gate.

---
_Generated by [Claude Code](https://claude.ai/code/session_01YXhNjBFWnM4SDAUxVV8hu3)_


backend/agents/investment_team/tests/test_predicate_ir_snapshot.py
```diff
@@ -0,0 +1,309 @@
...
+    return GROUP_SEP.join(render_predicate_group(g) for g in groups)
+
+
+def _live_renders() -> dict[str, str]:
+    return {case_id: _render_case(code) for case_id, code in CASES.items()}
+
+
+def _load_snapshots() -> dict[str, str]:
+    live = _live_renders()
+    if UPDATE_SNAPSHOTS or not SNAPSHOT_PATH.exists():
```
brandonkindred: **[MEDIUM] testing** — The snapshot regeneration logic in `_load_snapshots` writes the golden file whenever it is missing, regardless of the `UPDATE_PREDICATE_IR_SNAPSHOTS` environment variable. This bypasses the explicit opt-in safeguard and could allow a missing or accidentally deleted golden file to be silently regenerated with potentially incorrect current output, hiding regressions.

**Suggested fix:** Change the condition to only write when `UPDATE_PREDICATE_IR_SNAPSHOTS` is true. If the file is missing and the env var is not set, raise an error or fail the test with a clear message instructing the developer to set `UPDATE_PREDICATE_IR_SNAPSHOTS=1` to generate the initial snapshot. For example:
```python
if UPDATE_SNAPSHOTS:
    SNAPSHOT_PATH.parent.mkdir(parents=True, exist_ok=True)
    SNAPSHOT_PATH.write_text(...)
elif not SNAPSHOT_PATH.exists():
    raise FileNotFoundError(
        f"Snapshot file {SNAPSHOT_PATH} not found. "
        "Run with UPDATE_PREDICATE_IR_SNAPSHOTS=1 to generate it."
    )
```

backend/agents/investment_team/tests/test_predicate_ir_snapshot.py
```diff
@@ -0,0 +1,309 @@
...
+        return "<no groups>"
+    return GROUP_SEP.join(render_predicate_group(g) for g in groups)
+
+
+def _live_renders() -> dict[str, str]:
+    return {case_id: _render_case(code) for case_id, code in CASES.items()}
+
+
+def _load_snapshots() -> dict[str, str]:
+    live = _live_renders()
```
brandonkindred: **[LOW] standards** — `_live_renders()` is called twice at module load time: once inside `_load_snapshots` (line 144) and once to populate `_LIVE` (line 153). This duplicates the extraction work for all 20 strategies.

**Suggested fix:** Compute the live renders once and pass them to `_load_snapshots`, or restructure so that `_load_snapshots` does not call `_live_renders()` when not updating. For example, move the `_live_renders()` call outside and pass the result as an argument.

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
```diff
@@ -0,0 +1,463 @@
...
+def render_predicate_group(group: PredicateGroup) -> str:
+    """Render *group* (denylist + tree) as a stable canonical string.
+
+    Preconditions:
+        ``group`` is a :class:`PredicateGroup`.
+    Postconditions:
+        Returns ``"denied: {<sorted syms>}|none\\n<tree render>"`` —
+        deterministic for a given group.
+    """
+    if group.denied_symbols:
```
brandonkindred: **[LOW] logic** — `render_predicate_group` checks `if group.denied_symbols:` which treats an empty `frozenset` as falsy, causing it to render 'none' instead of an empty set. While an empty denylist is semantically equivalent to `None`, the rendering could be misleading if an empty frozenset is ever passed explicitly.

**Suggested fix:** Use an explicit `is not None` check: `if group.denied_symbols is not None:` to distinguish between `None` and an empty frozenset. This makes the intent clearer and avoids potential confusion.

brandonkindred (commented):
The PR cleanly extracts the BarPredicate IR into a dedicated, self-contained module with thorough documentation and a canonical renderer. The snapshot tests provide strong regression protection for the extractor's output across 20 representative strategies. The code adheres to SOLID principles and Design by Contract. Minor improvements are suggested for the snapshot regeneration logic and a few edge cases.

**Spec compliance:** The implementation fully meets the remaining acceptance criteria: the IR now lives in `coverage_probe/predicate_ir.py` as a standalone module, and snapshot tests pin the extracted IR shape. No logic changes were made to the moved code.

- `backend/agents/investment_team/tests/test_predicate_ir_snapshot.py` — **[INFO] testing** — Several pure utility functions (`_tree_and_unknown`, `_tree_or_unknown`, `_leg_gate_symbols`, `_find_or_groups`, `_tree_effective_symbols`) are not covered by direct unit tests in this file, though they are exercised by the broader test suite. Adding focused unit tests would strengthen the contract and make future refactors safer.

**Suggested fix:** Consider adding dedicated unit tests for these functions, similar to the existing tests for `_strip_symbol_gates` and `_eval_tree`, to ensure their contracts are independently verified.

backend/agents/investment_team/tests/test_predicate_ir_snapshot.py
brandonkindred: Fixed in fb022d4. `_load_snapshots` now writes the golden **only** when `UPDATE_PREDICATE_IR_SNAPSHOTS` is set; a missing golden without the flag raises `FileNotFoundError` with the regeneration instruction, so a deleted/absent golden fails loudly instead of being silently re-created from possibly-regressed output.

---

backend/agents/investment_team/tests/test_predicate_ir_snapshot.py
brandonkindred: Fixed in fb022d4. The live renders are now computed once into `_LIVE` and passed into `_load_snapshots(live)`, which no longer calls `_live_renders()` itself — the extraction pass runs a single time at module load.

---

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
brandonkindred: Fixed in fb022d4. `render_predicate_group` now uses `if group.denied_symbols is not None`, so an explicit (even empty) denylist renders as `denied: {}` while `None` renders as `denied: none` — the two are never conflated. Added a unit-test assertion locking the empty-frozenset case. Snapshot-neutral for the existing cases (none carry an empty denylist).

---

backend/agents/investment_team/strategy_lab/coverage_probe/indicator_probe.py
```diff
@@ -41,6 +41,25 @@
     LikelyBlocker,
     SubconditionCoverage,
 )
+from investment_team.strategy_lab.coverage_probe.predicate_ir import (
```
brandonkindred, line 44: **[MEDIUM] standards** — The import block imports several functions with leading underscores (e.g., _build_and_group, _collect_legs, _eval_tree, _find_or_groups, _leg_gate_symbols, _tree_and_unknown, _tree_effective_symbols, _tree_or_unknown) from predicate_ir.py. The underscore prefix conventionally marks these as private to the defining module, yet they are part of the stable contract between the extractor and the IR layer (as stated in predicate_ir.py's module docstring). This breaks encapsulation and may lead to accidental breakage if someone refactors predicate_ir.py assuming these are internal-only.

**Suggested fix:** Rename these functions to remove the leading underscore (e.g., build_and_group, collect_legs, eval_tree, etc.) in predicate_ir.py and update the import in indicator_probe.py accordingly. This clearly communicates that they are intended for external use and are part of the IR module's public API.

backend/agents/investment_team/tests/test_predicate_ir_snapshot.py
```diff
@@ -0,0 +1,380 @@
...
+        )
+    elif not SNAPSHOT_PATH.exists():
+        raise FileNotFoundError(
+            f"Snapshot file {SNAPSHOT_PATH} not found. "
+            "Run with UPDATE_PREDICATE_IR_SNAPSHOTS=1 to generate it."
+        )
+    return json.loads(SNAPSHOT_PATH.read_text(encoding="utf-8"))
+
+
+_LIVE = _live_renders()
```
brandonkindred: **[LOW] testing** — The module-level statement `_LIVE = _live_renders()` calls the extractor for all 20 test cases at import time. If the golden file is missing and UPDATE_PREDICATE_IR_SNAPSHOTS is not set, the module raises FileNotFoundError during import, which prevents test discovery entirely. Additionally, any import-time side effects of the extractor would execute during collection.

**Suggested fix:** Defer the computation to a pytest fixture (e.g., a session-scoped fixture that returns the live renders) or compute it lazily inside the test function. This ensures errors are raised during test execution rather than collection, and avoids unnecessary work when tests are not selected.

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
```diff
@@ -0,0 +1,465 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+# ---------------------------------------------------------------------------
+# BarPredicate IR nodes
+# ---------------------------------------------------------------------------
+
+
+@dataclass(frozen=True)
```
brandonkindred, line 48: **[LOW] standards** — MaskLeaf is a frozen dataclass with a field `evaluator: Callable[[pd.DataFrame], pd.Series]`. Callables are generally not hashable, which means instances of MaskLeaf (and any tree containing them) cannot be used in sets or as dictionary keys. While the current codebase does not require hashing, this is a latent pitfall for future consumers of the IR.

**Suggested fix:** Add a note in the MaskLeaf docstring warning that instances are not hashable due to the callable field. If hashing becomes necessary, consider storing a reference to a named function or using a wrapper that provides a stable hash.

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
```diff
@@ -0,0 +1,465 @@
...
+    tree: BarPredicate
+    denied_symbols: Optional[frozenset] = None
+
+
+# ---------------------------------------------------------------------------
+# Tree utilities
+# ---------------------------------------------------------------------------
+
+
+def _strip_symbol_gates(node: BarPredicate) -> Tuple[BarPredicate, Optional[frozenset]]:
```
brandonkindred: **[INFO] logic** — The function _strip_symbol_gates has no callers in the codebase outside of its unit test. The PR intentionally retains it as part of the IR utility surface. This is acceptable but worth noting for future cleanup.

**Suggested fix:** If the function is intended as a public utility, consider removing the leading underscore and documenting its use case. Otherwise, no action required.

brandonkindred (commented):
The PR cleanly extracts the BarPredicate IR into a dedicated, self-contained module with comprehensive snapshot and unit tests. Code quality is high: well-documented, follows Design by Contract, and achieves excellent test coverage. Minor issues exist around private function imports and test module-level execution, but no critical or high-severity problems were found.

**Spec compliance:** The PR fulfills the remaining acceptance criteria: the IR now lives in a standalone predicate_ir.py module with no dependencies on the extractor or aggregator, and snapshot tests pin the canonical IR render for 20 representative strategies. The implementation matches the described changes and all existing tests pass.

backend/agents/investment_team/strategy_lab/coverage_probe/indicator_probe.py
```diff
@@ -41,6 +41,25 @@
...
+    _build_and_group,
+    _build_or_group,
+    _collect_legs,
+    _eval_tree,
+    _find_or_groups,
+    _leg_gate_symbols,
+    _tree_and_unknown,
+    _tree_effective_symbols,
+    _tree_or_unknown,
+)
```
brandonkindred, line 62: **[MEDIUM] integration** — The import of `_strip_symbol_gates` from `predicate_ir` contradicts the PR description, which states the function is dead code with no callers tree-wide and was intentionally not re-imported into `indicator_probe.py`. If the function is truly unused in this file, the import should be removed to avoid dead imports and confusion. If it is actually used, the PR description should be corrected.

**Suggested fix:** Verify whether `_strip_symbol_gates` is called anywhere in `indicator_probe.py`. If not, remove it from the import statement. If it is used, update the PR description to accurately reflect its usage.

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
brandonkindred: **[INFO] standards** — The function `_strip_symbol_gates` is intentionally kept despite having no callers in the codebase. While covered by unit tests, dead code can be a maintenance burden and may confuse future developers.

**Suggested fix:** Add a comment above the function explaining why it is retained (e.g., `# Reserved for external consumers or future use; intentionally kept as part of the IR utility surface.`) to prevent accidental removal.

brandonkindred (commented):
The PR successfully extracts the BarPredicate IR into a dedicated, well-documented module with comprehensive snapshot and unit tests. Code quality is high, with clean separation of concerns and thorough testing. Minor issues: an apparently unused import in indicator_probe.py contradicts the PR description, and the test file performs heavy computation at module level. Otherwise, the implementation is production-ready.

**Spec compliance:** The code fully meets the acceptance criteria: the IR nodes and utilities are moved to predicate_ir.py, and snapshot tests pin the IR shape for 20 representative strategies. The module is self-contained and independently testable.

- `backend/agents/investment_team/tests/test_predicate_ir_snapshot.py` — **[LOW] testing** — The `_LIVE = _live_renders()` call at module level executes the extraction for all test cases at import time. This can slow down test collection and cause import-time failures if the extraction logic breaks, making debugging harder.

**Suggested fix:** Wrap the live renders computation in a session-scoped pytest fixture, e.g.:
```python
@pytest.fixture(scope="session")
def live_renders():
    return _live_renders()
```
Then use that fixture in the parametrized test and in `_load_snapshots`.

backend/agents/investment_team/strategy_lab/coverage_probe/indicator_probe.py
brandonkindred: Kept the underscore prefixes (addressed in f6eeb21 by clarifying the docstring rather than renaming). Rationale: this is one cohesive `coverage_probe` package — the helpers were already underscore-private in the original single `indicator_probe.py`, and the approved scope for this PR was a verbatim move with no renames. The genuinely *public* surface is the IR node types (the dataclasses + `BarPredicate` union + `PredicateGroup`) and the canonical renderers; the tree-walking helpers are shared implementation reached only by the sibling extractor/aggregator within the same package. I updated the module docstring to state exactly that, which removes the apparent contradiction (docstring "stable contract" vs underscored helpers). A package-wide rename of the API is a reasonable future cleanup but out of scope here and would touch ~100 call sites for marginal benefit.

---

backend/agents/investment_team/tests/test_predicate_ir_snapshot.py
brandonkindred: Fixed in f6eeb21. The live-render extraction and golden loading now live in session-scoped fixtures (`live_renders` / `snapshots`), so the work — and the missing-golden `FileNotFoundError` — happens at test run time rather than at import/collection. A broken extractor or absent golden now surfaces as a test failure, not a collection error, and the renders are computed only when these tests are selected.

---

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
brandonkindred: Clarified in f6eeb21 (docstring note). The premise here is actually inverted: `MaskLeaf` *is* hashable. A `@dataclass(frozen=True)` generates `__hash__` from its fields, and Python callables — ordinary functions, lambdas, and bound methods — are all hashable (by identity), so `MaskLeaf` instances and any tree containing them work fine in sets / as dict keys. I added a docstring note stating this (it holds as long as `evaluator` is hashable, which the values produced by the extractor always are). No behavioral change needed.

---

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
brandonkindred: Resolved by removing it. In f6eeb21 I deleted `_strip_symbol_gates` (and its unit test) entirely — it had no production callers, so rather than carry dead code or a justifying comment, it's gone. `predicate_ir.py` remains at 100% line coverage.

---

backend/agents/investment_team/strategy_lab/coverage_probe/indicator_probe.py
brandonkindred: No contradiction — `_strip_symbol_gates` was never imported into `indicator_probe.py`. The `predicate_ir` import block there (now lines ~44–61) imports the helpers the extractor/aggregator actually use; `_strip_symbol_gates` was deliberately excluded, exactly as the PR description stated. The only importer was the test module. As of f6eeb21 the function is removed entirely, so the question is moot.

---

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
brandonkindred: Addressed by removing the function outright in f6eeb21 rather than annotating it — no dead code to explain. Coverage stays at 100%.

---

backend/agents/investment_team/strategy_lab/coverage_probe/indicator_probe.py
brandonkindred: Update: reversed course on this per maintainer preference and **did the rename** in 0577cd3. The nine cross-module tree helpers are now public — `eval_tree`, `collect_legs`, `find_or_groups`, `tree_and_unknown`, `tree_or_unknown`, `tree_effective_symbols`, `leg_gate_symbols`, `build_and_group`, `build_or_group` — across `predicate_ir.py`, `indicator_probe.py`, and the IR tests. Only the `*_walk` recursion helpers remain private to the module, and the docstring now lists the public surface accordingly. Mechanical rename only; `predicate_ir.py` stays at 100% coverage and all probe + IR tests pass.

---

brandonkindred (commented):
The code successfully extracts the BarPredicate IR into a dedicated, well-documented module with comprehensive unit and snapshot tests. The implementation follows SOLID principles, uses Design by Contract, and achieves high test coverage. The only notable issues are minor documentation inconsistencies between the PR description and the actual code.

**Spec compliance:** The code meets the acceptance criteria: the IR is moved to a standalone module (`predicate_ir.py`), and snapshot tests are added to pin the IR shape. All specified cases are covered. The PR description mentions a `_strip_symbol_gates` function that is not present; this does not affect functionality but should be clarified.

- `backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py` — **[MEDIUM] spec-compliance** — The PR description states that the dead `_strip_symbol_gates` function is kept in the IR module and covered by a direct unit test. However, the function is not present in `predicate_ir.py`, and no test for it exists in `test_predicate_ir_snapshot.py`. This discrepancy could mislead reviewers and future maintainers about the module's contents.

**Suggested fix:** Either add the `_strip_symbol_gates` function to `predicate_ir.py` and a corresponding unit test in `test_predicate_ir_snapshot.py`, or update the PR description to accurately reflect that the function was removed. If the function is intentionally omitted, remove the mention from the PR description.
- `PR description` — **[LOW] documentation** — The PR description refers to functions with leading underscores (e.g., `_eval_tree`, `_collect_legs`, `_tree_and_unknown`), but the actual module exports them without underscores (e.g., `eval_tree`, `collect_legs`, `tree_and_unknown`). This inconsistency may cause confusion when cross-referencing.

**Suggested fix:** Update the PR description to use the correct public function names as exported by `predicate_ir.py`.

brandonkindred:
Addressed both documentation findings from the latest review by updating the PR description (the code was already correct):

- **`_strip_symbol_gates` discrepancy** — the function was intentionally removed during review (no callers tree-wide), so I dropped the stale "kept + unit-tested" claim and the description now states it was removed and is not present in `predicate_ir.py`. Chose the "update the description" option rather than re-adding dead code.
- **Underscore vs. public names** — the description now lists the helpers by their actual exported (public) names: `eval_tree`, `collect_legs`, `tree_and_unknown`/`tree_or_unknown`, `tree_effective_symbols`, `find_or_groups`, `leg_gate_symbols`, `build_and_group`/`build_or_group`. Only the `*_walk` recursion helpers remain private.

No code changes were needed for these two items; the description is now consistent with the module.

---

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
```diff
@@ -0,0 +1,466 @@
...
+    iteration order.
+
+    Preconditions:
+        ``node`` is a :data:`BarPredicate`; ``indent >= 0``.
+    Postconditions:
+        Returns a newline-joined string indented two spaces per level.
+        IR trees equal by shape, labels, ``unknown`` flags and gate
+        symbols render to byte-identical strings.
+    """
+    assert indent >= 0, "indent must be non-negative"
```
brandonkindred: **[LOW] standards** — The precondition check in `render_bar_predicate` uses a bare `assert indent >= 0`. In production environments where assertions may be disabled (e.g., `python -O`), this check would be skipped, potentially leading to confusing output or errors downstream. A proper `if`/`raise ValueError` is more robust.

**Suggested fix:** Replace the assert with an explicit check:
```python
if indent < 0:
    raise ValueError(f"indent must be non-negative, got {indent}")
```

backend/agents/investment_team/tests/test_predicate_ir_snapshot.py
```diff
@@ -0,0 +1,378 @@
...
+import difflib
+import json
+import os
+import textwrap
+from pathlib import Path
+
+import pandas as pd
+import pytest
+
+from investment_team.strategy_lab.coverage_probe.indicator_probe import _extract_subconditions
```
brandonkindred, line 35: **[LOW] testing** — The test imports `_extract_subconditions`, a private function from `indicator_probe`. While acceptable for white-box testing, it creates a dependency on an internal implementation detail that could change without notice, breaking the snapshot tests even if the IR contract is unchanged.

**Suggested fix:** Consider exposing a public test helper or making `_extract_subconditions` part of the public API if it is stable enough. Alternatively, document this dependency clearly so future maintainers are aware.

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
```diff
@@ -0,0 +1,466 @@
...
+    :class:`OrOp` (i.e. is an OR alternative rather than an AND-required
+    leg), and an integer identifier of the closest enclosing
+    :class:`OrOp` (used to group OR alternatives for the "all-zero OR"
+    blocker). Stops at each :class:`Leg` — does *not* descend into
+    ``Leg.inner``.
+    """
+    return _collect_legs_walk(node, accumulated_syms, in_or=False, or_id=None, next_or_id=[0])
+
+
+def _collect_legs_walk(
```
brandonkindred, line 212: **[INFO] standards** — The `_collect_legs_walk` and `_find_or_groups_walk` functions use a mutable list (`next_or_id`) to maintain state across recursive calls. This pattern works but is non-obvious and could be replaced with a simpler generator or an instance attribute if the walker were a class.

**Suggested fix:** No action required; this is an informational note. The current implementation is correct and performant.

brandonkindred (commented):
The PR successfully extracts the BarPredicate IR into a dedicated, well-documented module (`predicate_ir.py`) with a clean public API. Snapshot tests cover 20 representative strategies, and unit tests exercise all public helpers and renderer branches. The code follows SOLID principles, uses Design by Contract where appropriate, and integrates seamlessly with the existing `indicator_probe.py`. No critical or high-severity issues were found. A few low-severity improvements are suggested (assert vs. explicit raise, test dependency on private function) but do not block merge.

**Spec compliance:** The implementation fully satisfies the remaining acceptance criteria from #471: the IR now lives in a dedicated module, and snapshot tests pin the extracted IR shape. The removal of dead code (`_strip_symbol_gates`) and promotion of cross-module helpers to public names are positive side-effects. All specified verification steps (linting, CI, coverage) pass.

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
brandonkindred: Done in b195ab9. Replaced the bare `assert indent >= 0` in `render_bar_predicate` with an explicit `if indent < 0: raise ValueError(...)`, so the precondition holds even under `python -O`, and updated `test_render_indent_precondition` to expect `ValueError`. (The DbC convention here sanctions either an assert or explicit validation that raises — switched to the raise form for exactly the production-robustness reason you flag.) `predicate_ir.py` stays at 100% line coverage.

---

backend/agents/investment_team/tests/test_predicate_ir_snapshot.py
brandonkindred: Documented in b195ab9. Added a comment above the import making the dependency deliberate: this suite drives the extractor's entry point precisely so the golden render acts as a regression guard on its *output* — a change to `_extract_subconditions` that alters the IR shape **should** break these snapshots; that's the test's purpose, not accidental coupling to an implementation detail. I kept it white-box rather than promoting `_extract_subconditions` to public API, since exposing the extractor entry point is a broader API decision beyond this test's scope.

---

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
brandonkindred: Acknowledged — no change, per your note that none is required. The `next_or_id=[0]` single-element list is an intentional mutable cell threading a monotonic OR-group counter through the recursion (so sibling subtrees get distinct OR-group ids); converting to a walker class or a generator would be a larger refactor for no functional gain. Leaving the implementation as-is.

---

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
```diff
@@ -0,0 +1,467 @@
...
+    ``effective_unknown`` becomes ``AndOp.unknown`` — replaces the old
+    ``_Group.has_unknown_and_conjunct`` flag.
+    """
+    tree: BarPredicate = AndOp(legs=tuple(legs), unknown=effective_unknown)
+    if effective_symbols is not None:
+        tree = SymbolGate(syms=frozenset(effective_symbols), inner=tree)
+    return PredicateGroup(tree=tree, denied_symbols=denied_symbols)
+
+
+def build_or_group(
```
brandonkindred, line 387: **[LOW] testing** — The `build_or_group` function accepts an `ancestor_unknown` parameter that sets `AndOp.unknown` when ancestors are present. This flag is used in production (see `indicator_probe.py` line 1287-1290), but no snapshot or unit test covers a case where `ancestor_unknown=True`. Without a test, regressions in the rendered shape or tree structure for such groups could go undetected.

**Suggested fix:** Add a snapshot case that produces an OR group with AND-required ancestors where at least one ancestor conjunct is un-modellable (triggering `ancestor_unknown=True`). For example, a strategy like `if self.custom_ok(bar) and (close > 100 or close < 50): pass` would exercise this path. Alternatively, add a unit test for `build_or_group` with `ancestor_unknown=True` and verify the resulting `AndOp.unknown` flag and rendered output.

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
```diff
@@ -0,0 +1,467 @@
...
+            return True
+        return any(tree_or_unknown(leg) for leg in node.legs)
+    if isinstance(node, AndOp):
+        return any(tree_or_unknown(leg) for leg in node.legs)
+    if isinstance(node, (SymbolGate, Leg)):
+        return tree_or_unknown(node.inner)
+    return False
+
+
+def collect_legs(
```
brandonkindred, line 193: **[LOW] testing** — The `collect_legs` function exposes an `accumulated_syms` parameter in its public API, allowing callers to seed the symbol filter intersection. However, no test passes a non-None value for this parameter. If external code uses this parameter, its behavior is untested.

**Suggested fix:** Add a unit test that calls `collect_legs` with a non-None `accumulated_syms` and verifies that the returned legs have the correctly intersected symbol sets. For example, create a tree with a `SymbolGate` and pass an initial frozenset; assert the leg's effective syms are the intersection.

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
```diff
@@ -0,0 +1,467 @@
...
+    boolean :class:`pandas.Series` indexed by ``df.index``.
+
+    The :class:`SymbolGate` dispatch is the per-symbol filter: when
+    ``symbol`` is not in the gate's ``syms``, the entire sub-tree below
+    it evaluates to all-False without invoking the inner mask.
+    """
+    if isinstance(node, MaskLeaf):
+        try:
+            series = node.evaluator(df)
+        except Exception as exc:  # noqa: BLE001  # pragma: no cover — defensive
```
brandonkindred: **[LOW] testing** — The `eval_tree` function's `MaskLeaf` branch catches all exceptions and returns an all-False Series (lines 310-312). This defensive code is marked `# pragma: no cover` and has no test. While unlikely in normal operation, a test would ensure the error path behaves correctly (returns False, logs debug) and doesn't introduce a crash.

**Suggested fix:** Add a unit test that provides a `MaskLeaf` with an evaluator that raises an exception (e.g., `lambda df: 1/0`). Assert that `eval_tree` returns an all-False Series and optionally check that the debug log was emitted.

backend/agents/investment_team/tests/test_predicate_ir_snapshot.py
```diff
@@ -0,0 +1,382 @@
...
+from pathlib import Path
+
+import pandas as pd
+import pytest
+
+# White-box import on purpose: this suite drives the extractor's entry point so
+# the golden render is a regression guard on its *output*. If a change to
+# _extract_subconditions alters the IR shape these snapshots should break — that
+# is the test's job, not accidental coupling to an implementation detail.
+from investment_team.strategy_lab.coverage_probe.indicator_probe import _extract_subconditions
```
brandonkindred, line 35: **[INFO] integration** — The test file imports the private function `_extract_subconditions` from `indicator_probe` for white-box snapshot testing. This is intentional and documented in the module docstring, but it creates a coupling to an internal implementation detail. If the function is renamed or refactored, the snapshot tests will break, which is the intended guard.

**Suggested fix:** No action required; the docstring adequately explains the rationale. Ensure the module docstring remains if the import is ever questioned.

brandonkindred (commented):
The PR successfully extracts the BarPredicate IR into a dedicated, well-documented module (`predicate_ir.py`) with comprehensive snapshot and unit tests. The code is clean, follows SOLID principles and Design by Contract, and integrates correctly with the existing codebase. Minor test coverage gaps exist for a few edge cases and API parameters, but they do not pose production risks. Overall, the implementation meets the specified requirements and is production-ready.

**Spec compliance:** The implementation fully meets the specified requirements: the IR is moved to `coverage_probe/predicate_ir.py` as a standalone module depending only on pandas+stdlib; snapshot tests pin the canonical render for 20 representative strategies; regeneration is gated behind an environment variable; the cross-module helpers are promoted to public names; and the unused `_strip_symbol_gates` helper is removed. All acceptance criteria are satisfied.

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
```diff
@@ -0,0 +1,467 @@
...
+    union: set = set()
+    saw_universal = False
+    for _leg, syms, _in_or, _or_id in legs:
+        if syms is None:
+            saw_universal = True
+        else:
+            union |= syms
+    if saw_universal:
+        return None
+    return frozenset(union) if union else None
```
brandonkindred, line 278: **[INFO] logic** — In `tree_effective_symbols`, when all legs have symbol gates but the union of their symbols is empty (e.g., all gates are empty frozensets), the function returns `None` (universal) instead of an empty frozenset. This is because `frozenset(union) if union else None` returns `None` for an empty union. In practice, groups with empty symbol intersections are dropped before emission (as documented in `PredicateGroup`), so this path is unreachable. However, if the function were called on such a tree directly, the return value would be misleading.

**Suggested fix:** Consider returning `frozenset()` instead of `None` when `union` is empty and `saw_universal` is False, to accurately represent 'no possible symbols'. Alternatively, document this edge case explicitly in the docstring. No production impact.

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
brandonkindred: Added a focused unit test in 95253eb (`test_build_or_group_ancestor_unknown_sets_and_flag`): it builds an OR group with an un-modellable AND-required ancestor (`ancestor_unknown=True`) and asserts the root `AndOp.unknown` is set (and `tree_and_unknown` agrees) while the nested `OrOp` stays clean. Went with the direct unit test rather than a new snapshot strategy so the golden didn't need regenerating for one flag. `predicate_ir.py` stays at 100% line coverage.

---

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
brandonkindred: Added `test_collect_legs_seeded_accumulated_syms_intersects` in 95253eb: it seeds `collect_legs(tree, frozenset({"AAPL", "GOOG"}))` over a `SymbolGate({"AAPL", "MSFT"}, …)` and asserts the leg's effective syms are the intersection `{"AAPL"}`. The previously-untested non-`None` `accumulated_syms` path is now locked.

---

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
brandonkindred: Added `test_eval_tree_maskleaf_evaluator_exception_returns_all_false` in 95253eb: a `MaskLeaf` whose evaluator raises is asserted to fail closed to an all-False boolean Series and to emit the debug log. Since the branch is now genuinely exercised, I removed its `# pragma: no cover`. `predicate_ir.py` remains at 100% line coverage — now without relying on the pragma.

---

backend/agents/investment_team/tests/test_predicate_ir_snapshot.py
brandonkindred: Acknowledged — no change, per your note that none is required. The white-box import is intentional and documented (both in the module docstring and now in a comment above the import); leaving it as-is.

---

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
brandonkindred: Documented in 95253eb. Added an edge-case note to `tree_effective_symbols`' docstring: when every leg is symbol-gated but the union is empty (only reachable if all gates are empty frozensets — which the extractor collapses and the group is then dropped before emission), it returns `None` rather than an empty `frozenset`, and callers never observe it because such groups are pruned upstream. Chose the documentation option over flipping the return value, since changing an unreachable path's semantics carries more risk than this [INFO]/no-production-impact note warrants.

---

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
```diff
@@ -0,0 +1,474 @@
...
+    """
+    inner = leg.inner
+    if isinstance(inner, SymbolGate):
+        return inner.syms
+    return None
+
+
+def build_and_group(
+    legs: List[Leg],
+    effective_symbols: Optional[set],
```
brandonkindred: **[MEDIUM] standards** — The type hint for `effective_symbols` parameter in `build_and_group` (line 360) and `build_or_group` (line 382) is `Optional[set]`, an unparameterized generic. This reduces type-checking precision and does not convey that the set contains strings.

**Suggested fix:** Change the type hint to `Optional[Set[str]]` (import `Set` from `typing`). For `build_and_group`: `effective_symbols: Optional[Set[str]]`. For `build_or_group`: `effective_symbols: Optional[Set[str]]`.

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
```diff
@@ -0,0 +1,474 @@
...
+    boolean :class:`pandas.Series` indexed by ``df.index``.
+
+    The :class:`SymbolGate` dispatch is the per-symbol filter: when
+    ``symbol`` is not in the gate's ``syms``, the entire sub-tree below
+    it evaluates to all-False without invoking the inner mask.
+    """
+    if isinstance(node, MaskLeaf):
+        try:
+            series = node.evaluator(df)
+        except Exception as exc:  # noqa: BLE001
```
brandonkindred: **[LOW] logic** — `eval_tree` catches `Exception` broadly when invoking a `MaskLeaf` evaluator. While this prevents a single broken subcondition from crashing the whole probe, it may silently suppress unexpected errors (e.g., TypeError, AttributeError) that indicate bugs in the evaluator function. The current log only records the exception message, not the traceback.

**Suggested fix:** Consider logging the full traceback at DEBUG level using `logger.debug('subcondition %r failed on %s', node.label, symbol, exc_info=True)` to aid debugging without changing the fail-closed behavior. Alternatively, catch a narrower set of expected exceptions if the evaluator contract is well-defined.

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
```diff
@@ -0,0 +1,474 @@
...
+    """
+    return _collect_legs_walk(node, accumulated_syms, in_or=False, or_id=None, next_or_id=[0])
+
+
+def _collect_legs_walk(
+    node: BarPredicate,
+    accumulated_syms: Optional[frozenset],
+    in_or: bool,
+    or_id: Optional[int],
+    next_or_id: List[int],
```
brandonkindred, line 217: **[LOW] standards** — The `_collect_legs_walk` function uses a mutable list `next_or_id` to maintain an integer counter across recursive calls. This pattern, while functional, can be confusing and is prone to accidental sharing if the list is reused. Since the function is private, it's acceptable, but a clearer approach would improve maintainability.

**Suggested fix:** Refactor to use a simple integer accumulator returned from each recursive call, or encapsulate the counter in a small class. Alternatively, add a comment explaining the mutable list pattern for clarity.

backend/agents/investment_team/tests/test_predicate_ir_snapshot.py
```diff
@@ -0,0 +1,429 @@
...
+    tree = AndOp(legs=(outer_or, SymbolGate(frozenset({"AAPL"}), inner_or)))
+    groups = find_or_groups(tree)
+    # Two OR groups, id-ordered to match collect_legs' assignment (outer first).
+    assert [oid for oid, _ in groups] == [0, 1]
+    assert [op for _, op in groups] == [outer_or, inner_or]
+    # No OR anywhere → empty.
+    assert find_or_groups(AndOp(legs=(leg,))) == []
+
+
+def test_tree_effective_symbols_union_and_universal() -> None:
```
brandonkindred, line 374: **[INFO] testing** — The unit test `test_tree_effective_symbols_union_and_universal` does not cover the edge case where all symbol gates are empty frozensets, leading to an empty union and the function returning `None`. The docstring mentions this case is pruned upstream, but a direct unit test would ensure the function behaves correctly if that pruning ever changes.

**Suggested fix:** Add a test case: `tree = AndOp(legs=(SymbolGate(frozenset(), Leg('x', _mask('x', True))),))` and assert `tree_effective_symbols(tree) is None`.

backend/agents/investment_team/tests/test_predicate_ir_snapshot.py
```diff
@@ -0,0 +1,429 @@
...
+from pathlib import Path
+
+import pandas as pd
+import pytest
+
+# White-box import on purpose: this suite drives the extractor's entry point so
+# the golden render is a regression guard on its *output*. If a change to
+# _extract_subconditions alters the IR shape these snapshots should break — that
+# is the test's job, not accidental coupling to an implementation detail.
+from investment_team.strategy_lab.coverage_probe.indicator_probe import _extract_subconditions
```
brandonkindred, line 35: **[INFO] integration** — The snapshot test imports the private function `_extract_subconditions` from `indicator_probe`. This is intentional white-box testing to pin the extractor's output, but it creates a coupling to an implementation detail. If the extraction entry point is renamed or refactored, the test will break. This is acceptable for a regression guard, but worth noting.

**Suggested fix:** No action required; the module docstring already documents the rationale. Ensure that any future refactor of the extractor updates this test accordingly.

brandonkindred (commented):
The PR successfully extracts the BarPredicate IR into a dedicated, well-documented module (`predicate_ir.py`) with comprehensive snapshot and unit tests. Code quality is high, follows SOLID and Design by Contract, and integrates cleanly with the existing codebase. Minor improvements possible in type hints, exception logging, and test coverage for an edge case.

**Spec compliance:** All acceptance criteria met: IR moved to standalone module, snapshot tests added pinning canonical renders for 20 representative strategies, public API promoted (leading underscores removed), dead code (`_strip_symbol_gates`) removed, snapshot regeneration gated behind environment variable. The module is self-contained (depends only on pandas+stdlib) and independently testable.

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
brandonkindred: Done in 78f50d5. Both `build_and_group` and `build_or_group` now type `effective_symbols` as `Optional[Set[str]]` (`Set` imported from `typing`, matching the module's existing `List`/`Tuple`/`Optional` style).

---

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
brandonkindred: Done in 78f50d5. The fail-closed `MaskLeaf` handler now logs with `exc_info=True`, so the full traceback is captured at DEBUG (the underlying `TypeError`/`AttributeError` stays diagnosable) while behaviour is unchanged — still fails closed to an all-False Series. I kept the broad `except` because the evaluator is an arbitrary user-supplied closure with no narrow exception contract; narrowing it risks letting a genuinely-broken subcondition crash the whole probe. The unit test still asserts the all-False result and the log line.

---

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
brandonkindred: Done in 78f50d5. Added a comment in `collect_legs` explaining the `next_or_id=[0]` single-element mutable cell — ints are immutable, so a one-element list threads a monotonic OR-group counter through the recursion, giving each `OrOp` a distinct id shared by all its alternatives — plus a one-line cross-reference in `find_or_groups`. Kept the list-cell over a walker-class / returned-accumulator refactor so the diff stays minimal and the recursion shape is unchanged.

---

backend/agents/investment_team/tests/test_predicate_ir_snapshot.py
brandonkindred: Added `test_tree_effective_symbols_empty_gates_returns_none` in 78f50d5, using exactly your suggested tree (`AndOp(legs=(SymbolGate(frozenset(), Leg("x", …)),))`) and asserting `tree_effective_symbols(...) is None`. The documented edge case is now locked directly, so it stays correct even if the upstream pruning ever changes.

---

backend/agents/investment_team/tests/test_predicate_ir_snapshot.py
brandonkindred: Acknowledged — no action, per your note. The white-box import stays; both the module docstring and the inline comment above the import document the rationale, and any extractor refactor is expected to update this test.

---

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
```diff
@@ -0,0 +1,483 @@
...
+        return
+    if isinstance(node, AndOp):
+        for leg in node.legs:
+            _find_or_groups_walk(leg, out, next_or_id)
+        return
+    if isinstance(node, (SymbolGate, Leg)):
+        _find_or_groups_walk(node.inner, out, next_or_id)
+
+
+def eval_tree(node: BarPredicate, df: pd.DataFrame, symbol: str) -> pd.Series:
```
brandonkindred, line 312: **[LOW] standards** — The eval_tree function uses isinstance checks to dispatch on node types, which violates the Open/Closed principle. Adding a new BarPredicate variant requires modifying this function.

**Suggested fix:** Consider using a visitor pattern or a dispatch dictionary keyed by type to allow extension without modification. For example, define a method on each node class or use a registry. However, given the small number of node types, this is acceptable for now.

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
```diff
@@ -0,0 +1,483 @@
...
+    boolean :class:`pandas.Series` indexed by ``df.index``.
+
+    The :class:`SymbolGate` dispatch is the per-symbol filter: when
+    ``symbol`` is not in the gate's ``syms``, the entire sub-tree below
+    it evaluates to all-False without invoking the inner mask.
+    """
+    if isinstance(node, MaskLeaf):
+        try:
+            series = node.evaluator(df)
+        except Exception:  # noqa: BLE001
```
brandonkindred, line 323: **[LOW] logic** — The eval_tree function catches all exceptions (except Exception) when evaluating a MaskLeaf, which could suppress unexpected errors like MemoryError or SystemExit. While the intent is to fail closed, a broad except may hide critical bugs.

**Suggested fix:** Consider catching only specific expected exceptions (e.g., TypeError, ValueError, AttributeError) or at least re-raise system-exiting exceptions. Alternatively, ensure that the logging at DEBUG level with exc_info=True is sufficient for diagnosis.

backend/agents/investment_team/tests/test_predicate_ir_snapshot.py
```diff
@@ -0,0 +1,437 @@
...
```
brandonkindred, line 35: **[INFO] testing** — The test imports a private function (_extract_subconditions) from the indicator_probe module, which couples the test to implementation details. This is acceptable for snapshot tests but may require test updates if the internal function is refactored.

**Suggested fix:** No action required; this is a deliberate white-box test. Ensure that any refactoring of the extractor updates these tests accordingly.

brandonkindred (commented):
The PR successfully extracts the BarPredicate IR into a dedicated, well-documented module with comprehensive unit and snapshot tests. The code is clean, follows naming conventions, and integrates properly with the existing codebase. Minor design considerations (Open/Closed principle, broad exception handling) are noted but do not impact production readiness.

**Spec compliance:** The PR fully satisfies the remaining acceptance criteria: the IR is moved to a standalone module (predicate_ir.py) and snapshot tests are added to pin the IR shape. All specified cases are covered.

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
brandonkindred: By design, and I'll keep it as-is (you note it's acceptable here too). `BarPredicate` is a **closed** tagged union — the variants are fixed and few — so structural `isinstance` dispatch backed by an exhaustiveness guard (`raise AssertionError("unknown BarPredicate variant")`) is the idiomatic, type-checker-friendly shape. A visitor pattern or dispatch dict would scatter each algorithm's logic across the node classes (or a registry) and make `eval_tree`/the walkers harder to read, buying extensibility we deliberately don't want: adding a variant should be a conscious, test- and type-guarded change across the small set of readers, not a drop-in plugin. If the union ever grows materially, revisiting is cheap. No change.

---

backend/agents/investment_team/strategy_lab/coverage_probe/predicate_ir.py
brandonkindred: Kept deliberately, with one clarification on the premise: `except Exception` does **not** catch `SystemExit`, `KeyboardInterrupt`, or `GeneratorExit` — those derive from `BaseException`, not `Exception`, so they already propagate uninterrupted. `MemoryError` is the only system-ish case caught, and failing that one subcondition closed is the right behaviour: a single pathological evaluator shouldn't abort the whole coverage probe, and if memory is genuinely exhausted the next allocation fails anyway. The evaluator is an arbitrary user-supplied closure with no narrow exception contract, so enumerating "expected" types would be guesswork that risks silently letting a real bug through *or* crashing the run. As you note, the `exc_info=True` DEBUG logging added in 78f50d5 keeps any caught error fully diagnosable. No change.

---

backend/agents/investment_team/tests/test_predicate_ir_snapshot.py
brandonkindred: Acknowledged — no action, per your note; this is a deliberate white-box test. (As in the prior rounds: the rationale is documented both in the module docstring and the inline comment above the import, and any extractor refactor is expected to update this test.)

---

PR #828, opened 2026-06-08
Remove spec-unaware diversification warning from backtest anomaly gate

### LINKED ISSUE #827: Strategy Lab raises a spurious "no diversification" warning on single-asset strategies
## Problem

The Strategy Lab Quality Gates surface this warning in the UI for a strategy whose hypothesis intentionally targets a single asset (e.g. EURUSD only):

> All 43 trades are long on EURUSD=X — no diversification.

Flagging a deliberately single-asset strategy for "no diversification" is wrong — the narrow focus is the strategy's design, not a defect. A diversification warning should only be raised when the strategy's spec **specifically intended multiple assets** but the realized ledger concentrated on one (e.g. a spec naming 5 stocks that only ever trades one). A spec naming a single asset must never produce the warning.

## Root cause

The warning is produced by `_check_trade_diversification`, a rule in `BacktestAnomalyDetector` (`backend/agents/investment_team/strategy_lab/quality_gates/backtest_anomaly.py`). It fires whenever every trade is on the same side and same symbol, with **no awareness of the strategy's intended `target_symbols`**. It only ever emits a `warning` (never `critical`), so it never affected run control flow — it is purely a UI annotation.

## Resolution

Diversification breadth is already owned by the spec-aware `TargetSymbolCoverageGate.check_breadth`, which warns only when `spec.target_symbols` declares more than one intended symbol and the ledger traded just one (and is silent for single-target / empty-target / empty-ledger cases). Once gated on intent, the anomaly rule would be a strict subset of `check_breadth` — a guaranteed duplicate warning — so it is removed rather than made spec-aware. A regression test guards against re-introducing the spec-unaware rule.

brandonkindred:
## Problem

The Strategy Lab Quality Gates surfaced this warning in the UI for a strategy whose hypothesis intentionally targets a single asset (EURUSD only):

> All 43 trades are long on EURUSD=X — no diversification.

Flagging a deliberately single-asset strategy for "no diversification" is wrong — the narrow focus is the strategy's design, not a defect.

## Root cause

`BacktestAnomalyDetector._check_trade_diversification` (`backend/agents/investment_team/strategy_lab/quality_gates/backtest_anomaly.py`) fired whenever every trade was on the same side and same symbol, with **no view of the strategy's intended `target_symbols`**. The emitted text was an exact match for the warning the user saw. It only ever emitted a `warning` (never `critical`), so it never affected run control flow — it was purely a UI annotation.

## Fix

Diversification breadth is already owned by the spec-aware `TargetSymbolCoverageGate.check_breadth`, which warns **only** when `spec.target_symbols` declares more than one intended symbol and the realized ledger traded just one — and is silent for single-target, empty-target, and empty-ledger cases (so the EURUSD-only case is correctly silent). Once gated on intent, the anomaly rule would be a strict subset of `check_breadth` (a guaranteed duplicate warning), so it is **removed** rather than made spec-aware.

- Delete `_check_trade_diversification` and its entry in the `_RULES` tuple.
- No changes to `BacktestAnomalyCtx`, the `check(...)` call sites, or `check_breadth` — the latter already delivers the requested behavior.

## Tests

- The desired behavior is already locked in by existing `check_breadth` tests in `test_target_symbol_coverage.py` (single-target skip; multi-target-but-single-traded warns; empty-target / empty-ledger skip).
- Added one regression test in `test_backtest_anomaly_realism.py` asserting the anomaly detector emits no diversification warning for a single-symbol, single-side ledger — guarding against re-introduction.

## Verification

- `test_backtest_anomaly_realism.py`, `test_backtest_anomaly_dsr_aware.py`, `test_backtest_anomaly_zero_trade_diagnostics.py`, `test_target_symbol_coverage.py`, `test_liquidity_and_cost_stress.py`, `test_coverage_probe_orchestrator_stage.py`, `test_strategy_lab_zero_trade_repair.py`, `test_strategy_lab_walk_forward_integration.py` — all pass.
- `ruff check` + `ruff format --check` clean on both changed files.

Closes #827

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qe41h3xvjAD5WU9sRVfTB7)_


backend/agents/investment_team/tests/test_backtest_anomaly_realism.py
```diff
@@ -723,3 +723,39 @@ def test_lookahead_combined_rate_uses_entry_and_exit_observations() -> None:
...
     assert "40/40" in look[0].details
     assert look[0].severity == "critical"
+
+
+# ---------------------------------------------------------------------------
+# A single-asset, single-side ledger is not an anomaly
+# ---------------------------------------------------------------------------
+
+
+def test_single_symbol_single_side_ledger_emits_no_diversification_warning() -> None:
```
brandonkindred: **[LOW] naming** — The test function name 'test_single_symbol_single_side_ledger_emits_no_diversification_warning' is overly long (10 words). While descriptive, it exceeds the recommended 1-3 word name length for functions and may reduce readability.

**Suggested fix:** Shorten to something like 'test_no_diversification_warning_for_single_symbol' or 'test_single_asset_no_diversification_warning' while keeping the docstring for full context.

brandonkindred (commented):
The PR cleanly removes the spec-unaware diversification check from BacktestAnomalyDetector and adds a well-documented regression test. The change is minimal, correct, and aligns with the existing spec-aware TargetSymbolCoverageGate. No critical or high issues found.

**Spec compliance:** The implementation fully meets the acceptance criteria: the _check_trade_diversification method and its rule entry are removed, and the new test verifies that a single-asset, single-side ledger no longer triggers a diversification warning. Existing tests continue to pass, confirming no regression.

backend/agents/investment_team/tests/test_backtest_anomaly_realism.py
brandonkindred: Addressed in 3e07813 — renamed to `test_single_asset_no_diversification_warning`, keeping the docstring for full context. Tests still pass.

---

backend/agents/investment_team/tests/test_backtest_anomaly_realism.py
```diff
@@ -723,3 +723,39 @@ def test_lookahead_combined_rate_uses_entry_and_exit_observations() -> None:
...
+            exit_date=f"2024-{((i % 12) + 1):02d}-{((i % 27) + 2):02d}",
+            symbol="EURUSD=X",
+            side="long",
+        )
+        for i in range(20)
+    ]
+    detector = BacktestAnomalyDetector()
+    # No market_data → the look-ahead heuristic self-skips, isolating the
+    # remaining rules; the baseline metrics are anomaly-clean.
+    results = detector.check(_baseline_metrics(), trades)
```
brandonkindred, line 758: **[LOW] testing** — The test asserts that no result contains 'diversification' or 'trades are long', but if the detector returns an empty list (e.g., due to a bug that skips all checks), the assertion passes vacuously. This could mask a regression where the entire anomaly detection is broken.

**Suggested fix:** Add an assertion that results is not empty, or that at least one non-diversification check ran, to ensure the detector is actually executing. For example: `assert len(results) > 0, 'Expected at least one anomaly check to run'` before the substring check.

backend/agents/investment_team/strategy_lab/quality_gates/backtest_anomaly.py
```diff
@@ -503,7 +489,6 @@ def _check_cost_sensitivity(self, ctx: BacktestAnomalyCtx) -> Iterable[QualityGa
         _check_sharpe_ratio,
```
brandonkindred, line 496: **[INFO] standards** — The _RULES tuple no longer includes _check_trade_diversification, which is correct. However, there is no comment explaining why diversification is intentionally omitted, which could lead a future developer to re-add a similar spec-unaware check.

**Suggested fix:** Add a comment above the _RULES tuple: `# Diversification breadth is handled by TargetSymbolCoverageGate.check_breadth (spec-aware); do not add a spec-unaware rule here.`

brandonkindred (commented):
The PR cleanly removes the spec-unaware diversification warning from the backtest anomaly detector and adds a focused regression test. The changes are minimal, correct, and align with the acceptance criteria. No critical or high issues were found. Two minor suggestions are offered to strengthen the test and prevent future regressions.

**Spec compliance:** The implementation fully meets the spec: `_check_trade_diversification` is removed from the `_RULES` tuple (line 489-494) and its method definition is absent from the shown diff. The regression test `test_single_asset_no_diversification_warning` correctly verifies that a single-asset, single-side ledger produces no diversification warning. Existing tests pass, and no other components were modified.

backend/agents/investment_team/tests/test_backtest_anomaly_realism.py
brandonkindred: Good catch — added `assert results, "anomaly detector returned no results — checks did not run"` before the substring check (commit 3c17429). The detector's contract guarantees a non-empty list (a clean ledger yields the "passed all anomaly checks" info), so this now fails loudly if the detector ever silently stops running rather than passing vacuously.

---

backend/agents/investment_team/strategy_lab/quality_gates/backtest_anomaly.py
brandonkindred: Done (commit 3c17429) — added a comment above `_RULES` explaining that diversification breadth is intentionally owned by the spec-aware `TargetSymbolCoverageGate.check_breadth`, and that a spec-unaware rule must not be re-added here (it would false-positive on deliberately single-asset strategies).

---

PR #831, opened 2026-06-08
Code Review page: inline PR expansion + persisted per-PR review history

### LINKED ISSUE #830: Code Review page: expand PRs inline and show persisted per-PR review history
## Problem

On the Code Review page, selecting a PR replaces the whole PR list with a confirmation/detail panel, and only a single active review is shown at a time. There is no way to see the reviews that have already been executed.

## Desired behaviour

- Selecting a PR should **expand it inline within the list** (the other PRs stay visible), showing the PR detail currently shown on selection.
- After clicking **Start Review**, the review **status should appear on the PR's row** in the list.
- A PR that has been reviewed more than once should show, in its expanded view, a **table of all its reviews** with each review's status and outcome.
- The review history should **persist on the backend** so it survives page reloads and server restarts.

## Notes

Each review currently runs as a one-off `coding_team` job (`POST /review-pr`) with no per-PR index. Persistence requires a new `code_review_runs` table written by the coding-team service, a read endpoint, and a unified-API proxy, plus the frontend accordion/table and live-status hydration.

brandonkindred:
Closes #830

## What & why

Selecting a PR on the Code Review page used to replace the whole list with a confirmation/detail panel, and only one active review was shown at a time — and review history was lost entirely. Now:

- **Inline accordion** — selecting a PR expands it in place (other PRs stay visible) and shows the PR detail + **Start Review**. Re-clicking collapses it; only one PR is open at a time.
- **Live status on the row** — once a review starts, the PR row shows a status badge that updates live (running → outcome).
- **Per-PR reviews table** — the expanded view lists every review run on that PR with each run's status, outcome (event), finding counts, timestamp, and a link.
- **Backend persistence** — review history is stored server-side so it survives reloads and restarts.

## Backend

- **`code_review_runs` table** — new `coding_team` Postgres schema (`coding_team/postgres/__init__.py`) registered from a new app lifespan.
- **`review_history_store.py`** — best-effort `record_review_start` / `update_review` / `list_reviews`. Writes never block a review and no-op without Postgres; reads degrade to an empty list.
- **Lifecycle wiring** — `POST /review-pr` records the run; `_run_pr_review` keeps the row in lockstep with job state at the running and all terminal transitions (incl. failures, via `_record_failure`).
- **Read endpoints** — `coding_team` `GET /reviews` (runs for a repo, optionally one PR); unified API `GET /api/integrations/github/reviews` injects owner/repo from the GitHub config and proxies to the coding-team service.

## Frontend

- `CodeReviewPanelComponent` rewritten: per-PR `Map` of review records, accordion expansion, `Start Review`, row status badge, reviews table. Hydrates from the backend on load and **resumes live polling** for any still-running review. Multiple concurrent reviews across PRs are polled independently and torn down on destroy.
- New `CodeReviewRunItem` model + `IntegrationsApiService.getGitHubReviewHistory()`.

## Tests

- Store: live-Postgres suite (skipped without a DB, runs in CI) + always-on offline branch coverage.
- Review write-wiring + `GET /reviews` endpoint; unified-API proxy route (owner/repo injection, 503/502/504/error paths).
- Component: accordion, hydration, polling, badge branches, concurrency, teardown, and a rendered detail/table assertion (100% line coverage on the component); new service-method spec.

`ruff check`, `ng lint`, the frontend production build, and the affected backend/frontend test suites all pass locally.

## Verification

With the Docker stack up and GitHub configured: open Code Review → the list stays visible; expand a PR → start a review → watch the row status; run a second review → see both in the table; reload the page → history is restored from the backend.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HpuL8sQTm5fbZP1BCyGoyU)_


brandonkindred (commented):
The implementation is solid and meets the acceptance criteria. The inline accordion, live status badges, and persisted review history work as described. Testing is thorough across backend and frontend. The main areas for improvement are managing Angular subscriptions to prevent potential issues on component destruction, updating outdated CSS class names, and addressing minor backend code quality concerns.

**Spec compliance:** All specified features are implemented: inline PR expansion with accordion behavior, live status badge on PR rows, per-PR reviews table showing history, and backend persistence with graceful degradation. The code aligns with the project specification.

- `backend/unified_api/tests/test_integrations_github_pulls_route.py` — **[LOW] testing** — The test suite for GET /github/reviews does not cover the scenario where the coding team service returns a non-200 response with a non-JSON body. The production code handles this by falling back to resp.text[:300], but this path is untested.

**Suggested fix:** Add a test that simulates a 500 response with a plain text body and asserts that the proxy returns a 500 with the text as detail.

user-interface/src/app/components/code-review-panel/code-review-panel.component.html
```diff
@@ -159,19 +60,141 @@ <h3>Start a code review on this pull request?</h3>
...
+                </span>
+                <mat-icon class="cr-pull-row__arrow">
+                  {{ expandedPrNumber === pull.number ? 'expand_less' : 'chevron_right' }}
+                </mat-icon>
+              </button>
+
+              @if (expandedPrNumber === pull.number) {
+                <div class="cr-pull-detail">
+                  <!-- PR details -->
+                  <div class="cr-confirm-panel__pr">
```
brandonkindred: **[LOW] naming** — CSS class names like 'cr-confirm-panel__pr', 'cr-confirm-panel__meta', etc. reference the old confirmation panel design. Since the UI now uses inline expansion, these names are misleading and should be updated to reflect the current component structure (e.g., 'cr-pull-detail__pr').

**Suggested fix:** Rename the classes to align with the inline expansion pattern, such as 'cr-pull-detail__pr', 'cr-pull-detail__meta', 'cr-pull-detail__preview', 'cr-pull-detail__hint', 'cr-pull-detail__actions'. Update the corresponding SCSS rules accordingly.

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
```diff
@@ -59,21 +81,25 @@ export class CodeReviewPanelComponent implements OnInit, OnDestroy {
...
   ngOnInit(): void {
     this.checkGitHubConfig();
   }
 
   ngOnDestroy(): void {
-    this.stopPolling();
+    this.stopAllPollers();
   }
 
   checkGitHubConfig(): void {
```
brandonkindred, line 117: **[MEDIUM] standards** — The methods checkGitHubConfig (line 105), loadPulls (line 124), hydrateReviews (line 149), and startReview (line 206) subscribe to observables but do not store the subscriptions or unsubscribe on destroy. If the component is destroyed before an HTTP response arrives, the callbacks may still execute and attempt to update component state, potentially causing runtime errors. While Angular HttpClient observables complete, managing subscriptions is a best practice to prevent edge-case issues.

**Suggested fix:** Introduce a private destroy$ = new Subject<void>() and use takeUntil(this.destroy$) on each observable chain. Call this.destroy$.next() and this.destroy$.complete() in ngOnDestroy. Alternatively, store the Subscription objects and unsubscribe them in ngOnDestroy.

backend/agents/coding_team/review_history_store.py
```diff
@@ -0,0 +1,156 @@
...
+        sets.append("error = %s")
+        params.append(error)
+    if completed:
+        sets.append("completed_at = %s")
+        params.append(_now())
+    params.append(job_id)
+    try:
+        with get_conn() as conn, conn.cursor() as cur:
+            cur.execute(
+                f"UPDATE code_review_runs SET {', '.join(sets)} WHERE job_id = %s",
```
brandonkindred: **[LOW] security** — The update_review function constructs the SET clause using an f-string with string concatenation. Although the sets list currently contains only hardcoded strings, this pattern is fragile and could lead to SQL injection if future changes introduce user-controlled column names.

**Suggested fix:** Refactor to use a parameterized query with explicit column names, or validate each item in sets against an allowlist of known safe column names.

backend/agents/coding_team/api/main.py
```diff
@@ -746,10 +802,34 @@ def post_review_pr(request: ReviewPrRequest) -> ReviewPrResponse:
...
     _start_pr_review_thread(job_id, request, token)
     return ReviewPrResponse(job_id=job_id, pr_number=request.pr_number, pr_url=pr.html_url)
 
 
+@app.get("/reviews", response_model=List[ReviewRunItem])
+def get_reviews(
+    owner: str,
+    repo: str,
+    pr_number: Optional[int] = None,
+    limit: int = 500,
```
brandonkindred: **[INFO] standards** — The limit query parameter in get_reviews has no validation; it relies on the store to clamp the value. Adding validation at the API layer improves documentation and provides earlier feedback.

**Suggested fix:** Use limit: int = Query(default=500, ge=1, le=2000) to enforce the allowed range.

backend/agents/coding_team/api/main.py
```diff
@@ -697,6 +743,16 @@ def _build_review_code(files: List[Any]) -> ReviewCode:
     return ReviewCode("\n\n".join(blocks), reviewed, skipped)
 
 
+def _review_author() -> str:
+    """Resolve the author handle for a review row (best-effort, never raises)."""
+    try:
+        from agent_console.author import resolve_author
```
brandonkindred: **[INFO] standards** — The _review_author function imports resolve_author inside the function body. While this avoids import errors for an optional dependency, it is a minor code smell. Consider importing at the module level with a try-except or documenting the reason.

**Suggested fix:** Move the import to the top of the module inside a try-except block, or add a comment explaining why the inline import is necessary.

user-interface/src/app/components/code-review-panel/code-review-panel.component.html
brandonkindred: Done in 6624da1 — renamed `cr-confirm-panel__pr/meta/preview/hint/actions` to `cr-pull-detail__*` in the template and folded their rules into the `.cr-pull-detail` block (the obsolete `.cr-confirm-panel` block is gone).

---

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
brandonkindred: Done in 6624da1 — added a private `destroy$ = new Subject<void>()` and piped `takeUntil(this.destroy$)` onto all four HTTP subscriptions (`checkGitHubConfig`, `loadPulls`, `hydrateReviews`, `startReview`); `ngOnDestroy` now calls `destroy$.next()/complete()` before tearing down the pollers.

---

backend/agents/coding_team/review_history_store.py
brandonkindred: Done in 6624da1 — `update_review` now builds the SET clause from `(column, value)` pairs whose column names come only from the `_UPDATABLE_COLUMNS` allowlist, with an `assert` that the composed columns are a subset of it. The clause can no longer carry a caller-controlled identifier, and a future stray column fails loudly.

---

backend/agents/coding_team/api/main.py
brandonkindred: Done in 6624da1 — changed the parameter to `limit: int = Query(default=500, ge=1, le=2000)` and added a test asserting `limit=0` and `limit=3000` both return 422.

---

backend/agents/coding_team/api/main.py
brandonkindred: Addressed in 6624da1 by documenting it rather than moving it. The inline import is deliberate: author tagging is a best-effort convenience for persisted history, and keeping `agent_console` out of module load means a missing/broken `agent_console` (or its transitive deps) can never break importing this API or running a review. A module-level `try/except import` would execute that import at load time — the exact coupling this avoids — so I added a docstring note explaining the rationale instead.

---

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
```diff
@@ -124,61 +213,114 @@ export class CodeReviewPanelComponent implements OnInit, OnDestroy {
...
+          this.starting.delete(pull.number);
+          this.pullError = err?.error?.detail || err?.message || 'Failed to start review.';
+        },
+      });
   }
 
-  private startPolling(jobId: string): void {
-    this.stopPolling();
-    this.pollSub = pollJobStatus(
+  private startPolling(record: PrReviewRecord): void {
```
brandonkindred, line 302: **[MEDIUM] standards** — The startPolling method mutates PrReviewRecord objects in place without explicitly triggering Angular change detection. While the default change detection strategy and zone.js may pick up these changes, this approach is fragile and could break if the polling mechanism changes (e.g., running outside NgZone) or if the component is later switched to OnPush change detection. This could cause the live status badge and reviews table to stop updating in the UI.

**Suggested fix:** Inject ChangeDetectorRef and call this.cdr.markForCheck() after mutating the record properties in the polling callback. Alternatively, replace the record in the array with a new object reference (e.g., using spread operator) to ensure change detection detects the change. Example:

constructor(private cdr: ChangeDetectorRef) {}

// In startPolling callback:
record.status = status.status;
// ... other mutations
this.cdr.markForCheck();

backend/unified_api/routes/integrations.py
```diff
@@ -1649,3 +1663,66 @@ async def run_github_review_pr(body: RunPrReviewRequest) -> RunPrReviewResponse:
             status_code=502,
             detail=f"Coding team service returned an unexpected response: {e}",
         ) from e
+
+
+@router.get("/github/reviews", response_model=list[CodeReviewRunItem])
```
brandonkindred, line 1668: **[LOW] spec-compliance** — The GET /api/integrations/github/reviews endpoint does not expose the 'limit' query parameter that the underlying coding-team service supports. The frontend currently fetches all reviews (up to the default 500), but as review history grows, clients may want to control the number of results returned. This limits future scalability and API flexibility.

**Suggested fix:** Add an optional 'limit' query parameter to the endpoint and forward it to the coding-team service. Example:

limit: int = Query(default=500, ge=1, le=2000)
params: dict[str, Any] = {"owner": owner, "repo": repo, "limit": limit}
if pr_number is not None:
    params["pr_number"] = pr_number

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
```diff
@@ -124,61 +213,114 @@ export class CodeReviewPanelComponent implements OnInit, OnDestroy {
...
+            prUrl: resp.pr_url,
+          };
+          const list = this.reviews.get(pull.number) ?? [];
+          list.unshift(record); // newest-first
+          this.reviews.set(pull.number, list);
+          this.startPolling(record);
+        },
+        error: (err: { error?: { detail?: string }; message?: string }) => {
+          this.starting.delete(pull.number);
+          this.pullError = err?.error?.detail || err?.message || 'Failed to start review.';
```
brandonkindred: **[LOW] logic** — The component uses a single 'pullError' string to display errors from both pull request loading and review starting. If multiple operations fail (e.g., loading PRs fails and then starting a review also fails), only the last error message is shown, potentially hiding important context from the user.

**Suggested fix:** Consider using a more granular error state, such as a Map of errors keyed by PR number or separate properties for load errors vs. review start errors. At minimum, prepend new errors to the existing message or use an array of errors.

user-interface/src/app/components/code-review-panel/code-review-panel.component.spec.ts
```diff
@@ -154,28 +256,142 @@ describe('CodeReviewPanelComponent', () => {
...
+    expect(component['pollers'].size).toBe(2);
+
+    const callsBefore = apiSpy.getJobStatus.mock.calls.length;
+    component.ngOnDestroy();
+    expect(component['pollers'].size).toBe(0);
+    vi.advanceTimersByTime(15000);
+    expect(apiSpy.getJobStatus.mock.calls.length).toBe(callsBefore);
+  });
+
+  it('renders an expanded PR detail with a reviews table', async () => {
```
brandonkindred, line 451: **[INFO] testing** — The test suite verifies data changes during live polling but does not assert that the DOM updates accordingly (e.g., that the status badge text changes after a poll). While the 'renders an expanded PR detail' test checks initial DOM, adding a test that advances timers and inspects the rendered status text would increase confidence that the live UI updates work end-to-end.

**Suggested fix:** Add a test that starts a review, advances timers to simulate polling completion, calls fixture.detectChanges(), and then queries the DOM for the updated status badge or table cell text.

brandonkindred (commented):
The PR delivers a well-architected feature with clean separation of concerns, thorough backend persistence with graceful degradation, and comprehensive test coverage. The frontend component is well-structured with clear state management. Minor improvements are suggested for change detection robustness, API completeness, and error handling granularity.

**Spec compliance:** The implementation fully meets the acceptance criteria: inline accordion expansion with single-open behavior, live status badges on PR rows, per-PR review history table with all required columns, and backend persistence that survives reloads. All specified behaviors are implemented and tested.

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
```diff
@@ -59,58 +82,124 @@ export class CodeReviewPanelComponent implements OnInit, OnDestroy {
...
+        next: (items) => {
+          this.stopAllPollers();
+          const map = new Map<number, PrReviewRecord[]>();
+          for (const item of items) {
+            const record = this.toRecord(item);
+            const list = map.get(record.prNumber) ?? [];
+            list.push(record); // backend returns newest-first; preserve that order
+            map.set(record.prNumber, list);
+          }
+          this.reviews = map;
```
brandonkindred, line 208: **[HIGH] logic** — Race condition between hydrateReviews() and startReview(): When hydrateReviews() is in flight (e.g., during initial load or manual refresh), if the user starts a review via startReview(), the new review is added to the in-memory map and polling begins. When the hydrateReviews() response arrives, it calls stopAllPollers() (killing the new poller) and replaces this.reviews with the backend data, which may not include the newly started review (or includes it without the live poller). This causes the in-progress review to disappear from the UI or lose its live polling, leading to a broken user experience.

**Suggested fix:** Implement a merge strategy: before replacing this.reviews, preserve any records that have an active poller (i.e., non-terminal and not errored) and merge them into the new map. Alternatively, skip hydration if there are active pollers (this.pollers.size > 0) and defer until they complete, or cancel the hydration request if a start occurs. Example fix: in hydrateReviews() next callback, after building the new map, iterate over this.reviews and for any record that is still non-terminal and has a poller, ensure it is present in the new map (or keep the existing record). Also, do not call stopAllPollers() unconditionally; only stop pollers for jobs that are no longer present or are terminal in the new data.

backend/agents/coding_team/review_history_store.py
```diff
@@ -0,0 +1,166 @@
...
+    if status_text is not None:
+        assignments.append(("status_text", status_text))
+    if review_summary is not None:
+        assignments.append(("review_summary", Json(review_summary)))
+    if error is not None:
+        assignments.append(("error", error))
+    if completed:
+        assignments.append(("completed_at", _now()))
+    columns = [col for col, _ in assignments]
+    assert set(columns) <= _UPDATABLE_COLUMNS, f"non-allowlisted column(s): {set(columns) - _UPDATABLE_COLUMNS}"
```
brandonkindred: **[MEDIUM] standards** — The allowlist check for column names in the dynamically built UPDATE query uses an assert statement. In Python, assertions can be disabled globally (e.g., with -O flag), which would remove this validation. Although the column names are currently hardcoded and not user-controlled, relying on assert for a safety check is fragile and could lead to SQL injection if a future developer adds a user-controlled column name without realizing the assertion might be skipped.

**Suggested fix:** Replace the assert with an explicit if check that raises a ValueError or logs an error and returns early. For example:
if not set(columns) <= _UPDATABLE_COLUMNS:
    logger.error('non-allowlisted column(s): %s', set(columns) - _UPDATABLE_COLUMNS)
    return
This ensures the validation is always active regardless of optimization flags.

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
```diff
@@ -59,58 +82,124 @@ export class CodeReviewPanelComponent implements OnInit, OnDestroy {
...
+  }
+
+  /**
+   * Re-load persisted review history from the backend and rebuild the per-PR map,
+   * resuming a live poller for any review still in a non-terminal state. The
+   * backend is the source of truth, so this reconciles in-memory records with it.
+   * Best-effort: a failure leaves the page usable without history.
+   */
+  private hydrateReviews(): void {
+    this.integrationsApi
```
brandonkindred, line 177: **[LOW] performance** — The hydrateReviews() method calls getGitHubReviewHistory() without a pr_number parameter, fetching all review runs for the entire repository (up to 500). If the repository has many PRs and many review runs, this could load a large amount of data unnecessarily, especially since the UI only displays reviews for the currently visible PRs (paginated). This may cause performance issues and increased memory usage.

**Suggested fix:** Consider fetching reviews only for the currently loaded PRs by passing a list of pr_numbers (if the API supports multiple values) or fetching on demand when a PR is expanded. Alternatively, document the trade-off and ensure the backend limit is appropriate for typical use.

brandonkindred (commented):
The implementation largely meets the specification, with well-structured backend persistence, a clean frontend accordion, and thorough tests. However, a race condition in the frontend component can cause loss of in-progress review state when hydration overlaps with user actions, which must be fixed before merge. Additionally, the use of assert for security-sensitive validation in the store should be replaced with a robust check.

**Spec compliance:** All acceptance criteria are addressed: inline accordion, live row status, per-PR reviews table, backend persistence, hydration on load, and independent polling. The race condition undermines the reliability of the live review tracking, but the core functionality is present.

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
brandonkindred: Done in 4ae027a — injected `ChangeDetectorRef` and call `this.cdr.markForCheck()` after mutating the record in both the poll-status and connection-lost callbacks, so the live badge/table refresh independent of the change-detection strategy (safe under OnPush). Added a test that advances the poll timer, runs `detectChanges()`, and asserts the rendered row badge flips to `REQUEST_CHANGES` and the table status cell to `completed`.

---

backend/unified_api/routes/integrations.py
brandonkindred: Done in 4ae027a — added `limit: int = Query(default=500, ge=1, le=2000)` to the endpoint and forward it as `params["limit"]`. Covered with tests asserting the default (500) and an explicit value (10) are forwarded, and that out-of-range values (0, 5000) return 422 before any upstream call.

---

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
brandonkindred: Done in 4ae027a — a start-review failure now lands in a per-PR `reviewErrors: Map<number, string>` rendered inside that PR's expanded panel, while `pullError` is reserved for PR-list load failures. The two error surfaces are independent, so neither clobbers the other. Keying by PR number (rather than a flat array/prepend) is the natural fit since a start error always concerns one specific PR. Tests assert the per-PR error is set without touching `pullError` and that it renders in the expanded detail.

---

user-interface/src/app/components/code-review-panel/code-review-panel.component.spec.ts
brandonkindred: Done in 4ae027a — added `updates the rendered row badge + table as a live poll completes`: it starts a review, expands the PR, asserts the DOM badge shows `pending`, then advances the poll timer, runs `detectChanges()`, and asserts the badge flips to `REQUEST_CHANGES` and the reviews-table status cell to `completed` — verifying the live UI updates end-to-end.

---

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
brandonkindred: Good catch — fixed in ad06929. `hydrateReviews` no longer does `stopAllPollers()` + blanket replace. It now merges: it indexes every in-memory record that still has a live poller, rebuilds the map from the backend snapshot but **reuses the live record object** (same reference its poller mutates) for any matching job id, and **carries over** any still-polling review the snapshot didn't include yet (e.g. one started while the hydrate request was on the wire). Pollers are never killed; they self-terminate on terminal status / connection loss. Two new tests reproduce the race deterministically with a deferred hydrate `Subject`: one asserts an in-flight review absent from the snapshot survives with its poller, the other asserts the live record object (and its live status) is preferred over a stale snapshot copy.

---

backend/agents/coding_team/review_history_store.py
brandonkindred: Done in ad06929 — replaced the `assert` with an explicit guard: `unexpected = set(columns) - _UPDATABLE_COLUMNS; if unexpected: logger.error(...); return`. The check now stays active under `python -O`, and on a (currently impossible) non-allowlisted column it refuses to compose the query rather than proceeding. Added an offline test that shrinks `_UPDATABLE_COLUMNS` to force the branch and asserts `get_conn` is never called.

---

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
brandonkindred: Took the "document the trade-off" route (ad06929) — a per-PR-on-expand fetch can't work here because the **row status badges must show the latest review for every listed PR up front**, before any PR is expanded; lazy-loading would leave the list without badges. Fetching multiple specific PRs in one call isn't supported by the API (it filters by a single `pr_number`), and adding that is out of scope for this change. So hydration does one bounded call for the repo's recent reviews — now explicitly bounded by the `limit` param this PR added (default 500) — and I documented the rationale on `hydrateReviews`. For typical repos (tens of PRs, a handful of reviews each) this is well within the cap; if a repo outgrows it, the natural follow-up is server-side "latest run per PR" rather than per-PR fan-out from the client.

---

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
```diff
@@ -124,61 +244,125 @@ export class CodeReviewPanelComponent implements OnInit, OnDestroy {
...
+      (status: CodingTeamJobStatus) => {
+        // Mutate the record in place; markForCheck() makes the live badge/table
+        // refresh independent of the change-detection strategy (safe under OnPush).
+        record.status = status.status;
+        record.statusText = status.status_text;
+        record.reviewSummary = status.review_summary ?? record.reviewSummary;
+        record.prUrl = status.github_pr_url ?? record.prUrl;
+        record.error = status.error;
+        if (isCodingTeamTerminalStatus(status.status)) {
+          this.pollers.delete(record.jobId);
```
brandonkindred: **[MEDIUM] logic** — When a polled job reaches a terminal status, the subscription is removed from the `pollers` map but is not explicitly unsubscribed. If `pollJobStatus` does not complete its underlying observable, the subscription will continue to poll indefinitely, causing a memory leak and unnecessary network requests. The component relies on `pollJobStatus` to stop itself, but this contract is not enforced or tested.

**Suggested fix:** In the `onNext` callback, after detecting a terminal status, call `sub.unsubscribe()` before `this.pollers.delete(record.jobId)`. Alternatively, ensure `pollJobStatus` completes its observable and add a test that verifies no further `getJobStatus` calls occur after a terminal status is received. Example:
```typescript
if (isCodingTeamTerminalStatus(status.status)) {
  sub.unsubscribe();
  this.pollers.delete(record.jobId);
}
```

brandonkindred (commented):
The implementation is solid, well-tested, and meets all acceptance criteria. The backend persistence is robust with best-effort semantics, and the frontend component handles complex state (hydration, live polling, concurrency) correctly. A few minor improvements are suggested: ensure poller subscriptions are properly torn down on terminal status, add missing `markForCheck` for consistency, and clarify some code patterns. No critical or high-severity issues found.

**Spec compliance:** All specified features are implemented: inline accordion expansion, live row status badge, per-PR reviews table with status/outcome/findings/timestamp/link, and backend persistence via `code_review_runs`. The acceptance criteria are fully met. Tests cover the store, API endpoints, proxy, and component (100% line coverage claimed).

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
```diff
@@ -124,61 +244,125 @@ export class CodeReviewPanelComponent implements OnInit, OnDestroy {
...
+          const record: PrReviewRecord = {
+            jobId: resp.job_id,
+            prNumber: pull.number,
+            startedAt: Date.now(),
+            status: resp.status,
+            prUrl: resp.pr_url,
+          };
+          const list = this.reviews.get(pull.number) ?? [];
+          list.unshift(record); // newest-first
+          this.reviews.set(pull.number, list);
```
brandonkindred, line 274: **[LOW] standards** — After adding a new review record to the `reviews` map in `startReview`, `ChangeDetectorRef.markForCheck()` is not called. While the component currently uses default change detection and the HTTP callback triggers a cycle, this omission would break reactivity if the component is later switched to `OnPush` change detection. Other mutations in the component (e.g., polling updates) do call `markForCheck()`.

**Suggested fix:** Add `this.cdr.markForCheck();` after `this.reviews.set(pull.number, list);` to make the component safe under any change detection strategy and consistent with the rest of the code.

backend/agents/coding_team/api/main.py
```diff
@@ -858,7 +966,9 @@ def _run_pr_review(job_id: str, request: ReviewPrRequest, token: str) -> None:
             with GitHubClient(token=token) as client:
                 _record_failure(client, owner, repo, pr_number, job_id, f"code review failed: {e}")
         except Exception:  # noqa: BLE001 - the status update below is the last resort
-            update_job(job_id, status="failed", error=scrub_token_from_text(str(e)))
+            safe_err = scrub_token_from_text(str(e))
```
brandonkindred: **[LOW] logic** — In the nested exception handler inside `_run_pr_review`, the variable `e` from the outer `except` block is reused in `str(e)`. While this correctly references the original exception, it is confusing and could lead to mistakes if the inner except is later given its own exception variable.

**Suggested fix:** Rename the outer exception variable to `outer_exc` or add a comment clarifying that `e` still refers to the original review failure. Example:
```python
except Exception as review_exc:
    try:
        ...
    except Exception:
        safe_err = scrub_token_from_text(str(review_exc))
```

backend/agents/coding_team/review_history_store.py
```diff
@@ -0,0 +1,173 @@
...
+        # guard active even under ``python -O``.
+        logger.error("code_review_runs: refusing update with non-allowlisted column(s): %s", unexpected)
+        return
+    set_clause = ", ".join(f"{col} = %s" for col in columns)
+    params: list[Any] = [val for _, val in assignments]
+    params.append(job_id)
+    try:
+        with get_conn() as conn, conn.cursor() as cur:
+            cur.execute(
+                f"UPDATE code_review_runs SET {set_clause} WHERE job_id = %s",
```
brandonkindred: **[INFO] standards** — The `update_review` function composes the SET clause using an f-string: `f"UPDATE code_review_runs SET {set_clause} WHERE job_id = %s"`. Although the column names are restricted by an allowlist and a guard prevents unlisted columns, using string interpolation for SQL identifiers is a pattern that security scanners may flag. The current implementation is safe, but a more idiomatic approach would use `psycopg.sql.SQL` and `Identifier`.

**Suggested fix:** Consider replacing the f-string with `psycopg.sql.SQL` and `sql.Identifier` for the column names to make the safety structurally evident. Example:
```python
from psycopg import sql
identifiers = [sql.Identifier(col) for col in columns]
set_clause = sql.SQL(', ').join(sql.SQL('{} = %s').format(id) for id in identifiers)
query = sql.SQL('UPDATE code_review_runs SET {set_clause} WHERE job_id = %s').format(set_clause=set_clause)
```

backend/agents/coding_team/api/main.py
```diff
@@ -697,6 +743,22 @@ def _build_review_code(files: List[Any]) -> ReviewCode:
...
+def _review_author() -> str:
+    """Resolve the author handle for a review row (best-effort, never raises).
+
+    The ``agent_console`` import is intentionally inline rather than module-level:
+    author tagging is an optional convenience for persisted history, and keeping
+    the import out of module load means a missing/broken ``agent_console`` (or its
+    transitive deps) can never break importing this API or running a review.
+    """
+    try:
+        from agent_console.author import resolve_author
```
brandonkindred: **[INFO] standards** — The `_review_author` function imports `agent_console.author` inline inside a try/except. While this prevents import errors from breaking the API, it is unconventional and scatters import logic. A module-level import with a similar try/except would be cleaner and still safe.

**Suggested fix:** Move the import to the module level, wrapped in a try/except that sets a sentinel (e.g., `_resolve_author = None`) and handle the fallback in `_review_author`. This keeps imports grouped and the function simpler.

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
```diff
@@ -124,61 +244,125 @@ export class CodeReviewPanelComponent implements OnInit, OnDestroy {
...
-    });
+  /** The "Start Review" error for a PR, if its last attempt failed. */
+  reviewErrorFor(prNumber: number): string | null {
+    return this.reviewErrors.get(prNumber) ?? null;
   }
 
-  private startPolling(jobId: string): void {
-    this.stopPolling();
-    this.pollSub = pollJobStatus(
+  private startPolling(record: PrReviewRecord): void {
```
brandonkindred, line 302: **[INFO] documentation** — The `startPolling` method lacks a JSDoc comment explaining its contract: it mutates the given `PrReviewRecord` in place, relies on `pollJobStatus` to stop on terminal status, and registers the subscription for teardown.

**Suggested fix:** Add a JSDoc comment above `startPolling` describing its behavior and assumptions, e.g.:
```typescript
/**
 * Begin polling the job status for a review record. Mutates the record in place
 * on each status update. The poller is automatically torn down when the job
 * reaches a terminal state or on component destroy.
 */
```

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
```diff
@@ -52,65 +76,161 @@ export class CodeReviewPanelComponent implements OnInit, OnDestroy {
...
+   *
+   * Note: this fetches the repository's recent reviews in one call (bounded by
+   * the backend `limit`) rather than per-PR, because the row status badges need
+   * the latest review for *every* listed PR up front; a per-PR-on-expand fetch
+   * would leave the list without badges. Best-effort: a failure leaves the page
+   * usable without history.
+   */
+  private hydrateReviews(): void {
+    this.integrationsApi
+      .getGitHubReviewHistory()
```
brandonkindred, line 178: **[INFO] logic** — `hydrateReviews` calls `getGitHubReviewHistory()` without a `prNumber` or `limit` parameter, relying on the backend default limit of 500. If the repository accumulates more than 500 reviews, the latest review for some less active PRs may not be included, causing missing row badges or incomplete history. This is a conscious design choice, but it is not documented.

**Suggested fix:** Add a comment in `hydrateReviews` or the component's class JSDoc noting that history is limited to the most recent 500 reviews and that older reviews may not appear. If needed, a future enhancement could fetch per-PR on expansion.

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
brandonkindred: Done in 8b42bfa — added a `disposePoller(jobId)` helper that `unsubscribe()`s the subscription and removes it from the map, called on both the terminal-status and connection-lost paths (instead of only deleting from the map). `pollJobStatus` already completes its stream on terminal via an inclusive `takeWhile`, but the explicit unsubscribe means the component no longer relies on that contract. Added a test that, after a terminal poll, advances timers further and asserts `getJobStatus` is not called again.

---

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
brandonkindred: Done in 8b42bfa — added `this.cdr.markForCheck()` after recording the new review (and on the start-error path), consistent with the polling callbacks and safe under OnPush.

---

backend/agents/coding_team/api/main.py
brandonkindred: Done in 8b42bfa — renamed the outer `except Exception as e` to `review_exc` throughout the block and added a comment noting the inner `except` has no exception of its own, so `review_exc` is unambiguously the original review failure.

---

backend/agents/coding_team/review_history_store.py
brandonkindred: Done in 8b42bfa — `update_review` now composes the SET clause with `psycopg.sql`: `sql.SQL(", ").join(sql.SQL("{} = %s").format(sql.Identifier(col)) for col in columns)`, and the full statement via `sql.SQL("UPDATE code_review_runs SET {set_clause} WHERE job_id = %s").format(...)`. Column names are quoted identifiers (never string-interpolated) and values stay parameterized, so the safety is structurally evident — scanners won't flag it. Kept the allowlist guard as defense-in-depth.

---

backend/agents/coding_team/api/main.py
brandonkindred: Done in 8b42bfa — adopted the sentinel pattern: `try: from agent_console.author import resolve_author as _resolve_author except Exception: _resolve_author = None` at module level, with `_review_author()` returning `"anonymous"` when the sentinel is `None` or the call raises. Imports are now grouped and the function is simpler, while a missing/broken `agent_console` still can't break importing this API.

---

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
brandonkindred: Done in 8b42bfa — added a JSDoc contract to `startPolling` describing that it mutates the record in place (and `markForCheck()`s), registers the subscription in `pollers` for teardown, and explicitly disposes the poller on terminal status / connection loss (with `ngOnDestroy` tearing down any still-running ones).

---

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
brandonkindred: Done in 8b42bfa — expanded the `hydrateReviews` doc to state the cap explicitly: history is the most recent runs up to the backend `limit` (default 500), and in a repo with more than ~500 recent runs the oldest runs (and badges for the least-active PRs) may be absent until a future server-side "latest run per PR" query lands.

---

brandonkindred (commented):
The implementation successfully delivers the requested features: inline accordion expansion, live status badges, per-PR review history table, and backend persistence. The code is well-structured, follows naming conventions, has comprehensive test coverage, and includes proper error handling. A few minor improvements could be made, such as avoiding a variable name conflict and clarifying a fallback error handler, but overall the code is production-ready.

**Spec compliance:** The code meets all specified acceptance criteria: inline accordion with single-open behavior, live status badge on PR rows, per-PR reviews table with status/outcome/findings/timestamp/link, backend persistence via code_review_runs table, best-effort store that degrades gracefully, unified API proxy, and frontend hydration with polling resumption. All described verification steps are supported.

backend/agents/coding_team/review_history_store.py
```diff
@@ -0,0 +1,175 @@
...
+    Preconditions:
+        - ``owner``/``repo`` name the configured repository.
+    Postconditions:
+        - Returns up to ``limit`` rows ordered by ``created_at`` DESC. Returns
+          ``[]`` when Postgres is unavailable or the query fails (never raises).
+    """
+    if not is_postgres_enabled():
+        return []
+    limit = max(1, min(limit, 2000))
+    sql = (
```
brandonkindred: **[LOW] naming** — The local variable `sql` shadows the imported `psycopg.sql` module. This is confusing and could lead to errors if someone later tries to use `sql.Identifier` inside this function.

**Suggested fix:** Rename the local variable to `query` or `stmt` to avoid shadowing the module.

backend/agents/coding_team/api/main.py
```diff
@@ -830,35 +938,48 @@ def _run_pr_review(job_id: str, request: ReviewPrRequest, token: str) -> None:
...
+        logger.exception("PR review hook failed: %s", review_exc)
         try:
             with GitHubClient(token=token) as client:
-                _record_failure(client, owner, repo, pr_number, job_id, f"code review failed: {e}")
+                _record_failure(client, owner, repo, pr_number, job_id, f"code review failed: {review_exc}")
         except Exception:  # noqa: BLE001 - the status update below is the last resort
-            update_job(job_id, status="failed", error=scrub_token_from_text(str(e)))
+            # ``review_exc`` is the original review failure (the inner except has no
+            # exception of its own); surface it on both the job and the review row.
+            safe_err = scrub_token_from_text(str(review_exc))
```
brandonkindred, line 985: **[LOW] logic** — The fallback in the except block calls `update_job` and `update_review` even though `_record_failure` (called earlier) may have already succeeded in updating them. While the updates are idempotent, the code could be clarified to avoid confusion.

**Suggested fix:** Add a comment explaining that the fallback is a safety net and that duplicate updates are harmless.

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
```diff
@@ -52,65 +76,163 @@ export class CodeReviewPanelComponent implements OnInit, OnDestroy {
...
+   *
+   * Note: this fetches the repository's recent reviews in one call (the backend
+   * `limit`, default 500) rather than per-PR, because the row status badges need
+   * the latest review for *every* listed PR up front; a per-PR-on-expand fetch
+   * would leave the list without badges. Consequence of the cap: in a repo with
+   * more than ~500 recent runs, the oldest runs (and the badges for the least
+   * active PRs) may be absent until a future "latest run per PR" backend query
+   * lands. Best-effort: a failure leaves the page usable without history.
+   */
+  private hydrateReviews(): void {
```
brandonkindred, line 176: **[LOW] performance** — The `hydrateReviews` method fetches all reviews for the repository (up to 500) in a single request. For repositories with many PRs and review runs, this payload could become large and slow down the initial page load. The comment acknowledges the limitation, but it may impact user experience.

**Suggested fix:** Consider implementing pagination or fetching reviews only for visible PRs, or adding a backend endpoint that returns the latest review per PR to reduce payload size. This is acceptable for now but should be monitored.

backend/agents/coding_team/review_history_store.py
brandonkindred: Good catch — fixed in 996a300. Renamed the local query string in `list_reviews` from `sql` to `query` so it no longer shadows the imported `psycopg.sql` module (with a comment noting why).

---

backend/agents/coding_team/api/main.py
brandonkindred: Done in 996a300 — added a comment explaining the last-resort `except` is a safety net: `_record_failure` may already have marked the job/review failed, but if it itself raised (e.g. the GitHub client failed) these direct `update_job`/`update_review` writes ensure the job never wedges in "running", and both writes are idempotent so a duplicate update is harmless.

---

user-interface/src/app/components/code-review-panel/code-review-panel.component.ts
brandonkindred: Acknowledged — leaving the single bounded fetch as-is for this PR, which the comment already documents (and you note is acceptable for now). The constraint that makes it the right call here: row badges need the latest review for every listed PR up front, so a single repo-scoped call (capped by the `limit`, default 500) is the simplest correct source for them. The flagged scaling path — a backend "latest review per PR" endpoint — is the documented follow-up; I'd rather land that as its own change than expand this PR's surface. Noted for monitoring.

---

brandonkindred (commented):
The code is well-structured, follows best practices, and meets all acceptance criteria. The backend persistence is robust with best-effort semantics, and the frontend component handles concurrency and hydration correctly. Test coverage is thorough. One minor naming conflict in the review history store should be addressed to prevent future confusion.

**Spec compliance:** All specified features are implemented: inline accordion expansion, live status badges, per-PR reviews table, backend persistence with code_review_runs table, and review history that survives reloads. The implementation matches the PR description and closes #830.

backend/agents/coding_team/review_history_store.py
brandonkindred: **[MEDIUM] naming** — The variable `sql` in `list_reviews` shadows the imported `sql` module from `psycopg`. This can cause confusion and potential bugs if someone later tries to use `sql.Identifier` or other module attributes inside this function, because `sql` will be a string, not the module.

**Suggested fix:** Rename the variable to `query` or `stmt` to avoid shadowing the import. For example:
```python
query = (
    "SELECT job_id, owner, repo, pr_number, pr_url, status, status_text, "
    "       review_summary, error, author, created_at, completed_at "
    "FROM code_review_runs WHERE owner = %s AND repo = %s"
)
params: list[Any] = [owner, repo]
if pr_number is not None:
    query += " AND pr_number = %s"
    params.append(pr_number)
query += " ORDER BY created_at DESC LIMIT %s"
params.append(limit)
```

brandonkindred: Already addressed — this duplicates the earlier `sql`-shadowing comment and was fixed in 996a300 (which is the current HEAD of this branch). `list_reviews` now builds the query in a local named `query` (lines 159–172); the only `sql` reference left in the module is the imported `psycopg.sql` used by `update_review`. No remaining shadowing. Resolving.

---

backend/unified_api/routes/integrations.py
```diff
@@ -1649,3 +1663,68 @@ async def run_github_review_pr(body: RunPrReviewRequest) -> RunPrReviewResponse:
...
+            detail="Coding team service timed out while listing reviews.",
+        ) from e
+    except httpx.HTTPError as e:
+        logger.warning("github reviews: cannot reach coding team service at %s: %s", coding_team_url, e)
+        raise HTTPException(
+            status_code=502,
+            detail=f"Could not reach coding team service: {e}",
+        ) from e
+
+    if resp.status_code != 200:
```
brandonkindred, line 1716: **[MEDIUM] standards** — The `list_github_reviews` endpoint forwards upstream error detail directly to the frontend without sanitization. In the block at lines 1716-1721, `detail = resp.json().get("detail", resp.text[:300])` is used as the HTTPException detail. If the coding-team service returns a 500 error containing a stack trace or sensitive internal information, that data will be exposed to the frontend, creating an information disclosure risk.

**Suggested fix:** Sanitize the error detail before returning it, or use a generic message. For example:
```python
if resp.status_code != 200:
    try:
        upstream_detail = resp.json().get("detail", resp.text[:300])
    except Exception:
        upstream_detail = resp.text[:300]
    logger.warning("coding team reviews upstream error: %s", upstream_detail)
    raise HTTPException(status_code=502, detail="Failed to retrieve review history.")
```
This logs the real error internally while preventing accidental exposure to the client.

user-interface/src/app/components/code-review-panel/code-review-panel.component.spec.ts
```diff
@@ -125,57 +152,347 @@ describe('CodeReviewPanelComponent', () => {
...
+  it('ignores a second startReview while one is already starting', async () => {
     await setup();
-    component.confirmAndReview();
+    component.starting.add(1);
+    component.startReview(component.pulls[0]);
     expect(integrationsSpy.runGitHubReviewPr).not.toHaveBeenCalled();
   });
 
-  it('surfaces an error when starting the review fails', async () => {
+  it('surfaces a start-review error per PR without touching the list-load banner', async () => {
```
brandonkindred, line 316: **[LOW] testing** — The `startReview` error handler builds the error message as `err?.error?.detail || err?.message || 'Failed to start review.'`. The existing test only covers the `err.error.detail` path. The fallback to `err.message` (when `err.error` is absent) and the default string (when both are absent) are not tested, leaving two error-display branches uncovered.

**Suggested fix:** Add two additional test cases:
1. An error with `{ message: 'Network error' }` (no `error` property) to verify the `err.message` fallback.
2. An error with `{}` (empty object) to verify the default `'Failed to start review.'` string.
Example:
```typescript
it('surfaces a start-review error using err.message when detail is absent', async () => {
  integrationsSpy.runGitHubReviewPr.mockReturnValue(throwError(() => ({ message: 'Network down' })));
  await setup();
  component.startReview(component.pulls[0]);
  expect(component.reviewErrorFor(1)).toBe('Network down');
});
it('surfaces a default start-review error when no detail or message', async () => {
  integrationsSpy.runGitHubReviewPr.mockReturnValue(throwError(() => ({})));
  await setup();
  component.startReview(component.pulls[0]);
  expect(component.reviewErrorFor(1)).toBe('Failed to start review.');
});
```

brandonkindred (commented):
The PR delivers a well-engineered feature with strong adherence to Design by Contract, SOLID principles, and thorough testing. Backend persistence is robust (best-effort, never blocks), the store is secure (parameterized queries, column allowlist), and the frontend component handles concurrency, hydration races, and teardown correctly. Minor issues noted: a potential information disclosure in the unified API error propagation, possibly missing SCSS styles for status badges, and a small test coverage gap for error fallback paths. None are blocking.

**Spec compliance:** All acceptance criteria are met: inline accordion expansion, live row status badge, per-PR reviews table with status/outcome/findings/timestamp/link, backend persistence surviving reloads, hydration on load with live polling resumption, and independent concurrent review polling with proper teardown. The implementation aligns fully with the task specification.

- `user-interface/src/app/components/code-review-panel/code-review-panel.component.scss` — **[LOW] structure** — The component template uses CSS classes `cr-job-status`, `cr-job-status--completed`, `cr-job-status--failed`, and dynamic `cr-job-status--{{ rec.status }}`. The provided SCSS snippet (starting at line 92) does not include definitions for these classes. If they are not defined in the omitted lines 1–91 or in a global stylesheet, the status badges and table cells will lack proper styling, resulting in a degraded user experience.

**Suggested fix:** Verify that the SCSS file contains appropriate styles for `.cr-job-status` and its modifier classes. If missing, add them. Example:
```scss
.cr-job-status {
  display: inline-block;
  padding: 0.1rem 0.5rem;
  border-radius: 999px;
  font-size: 0.75rem;
  font-weight: 500;
  &--completed { background: #e8f5e9; color: #2e7d32; }
  &--failed { background: #ffebee; color: #c62828; }
  &--running { background: #e3f2fd; color: #1565c0; }
}
```

backend/unified_api/routes/integrations.py
brandonkindred: Done in 6797b8e — the `GET /github/reviews` proxy now logs the upstream detail server-side (`logger.warning(... resp.status_code, upstream_detail)`) and raises a generic `"Failed to retrieve review history."` to the client. I preserved the upstream status code (rather than collapsing to 502) so the frontend still distinguishes e.g. a 404 from a 500, while no internal body text is exposed. Updated the two upstream-error tests to assert the generic message.

Note: the sibling integrations proxies (`run_github_review_pr`, the issue/pulls list routes) use the same propagate-detail pattern — I scoped this fix to the flagged reviews endpoint to keep the PR focused; giving the others the same treatment would be a good consistency follow-up.

---

user-interface/src/app/components/code-review-panel/code-review-panel.component.spec.ts
brandonkindred: Done in 6797b8e — added both cases: one with `{ message: 'Network down' }` asserting `reviewErrorFor(1) === 'Network down'` (the `err.message` fallback), and one with `{}` asserting the default `'Failed to start review.'`. Component branch coverage went from 95% to ~98%.

---

PR #834, opened 2026-06-09
Strategy Lab: forward the strands transport timeout instead of dropping it

### LINKED ISSUE #833: Strategy Lab: strands transport timeout is silently dropped (Invalid configuration parameters: ['timeout'])
## Symptom

Strategy Lab strategy generation logs a warning on every model construction, and the transport read timeout is never actually applied:

```
investment_team.strategy_lab.agents.model_factory: Strands model: Ollama model_id=deepseek-v4-pro:cloud host=https://ollama.com cloud=True timeout=900s schema=True

model_factory.py:176: UserWarning: Invalid configuration parameters: ['timeout'].
Valid parameters are: ['additional_args', 'context_window_limit', 'keep_alive', 'max_tokens', 'model_id', 'options', 'stop_sequences', 'temperature', 'top_p'].
See https://github.com/strands-agents/sdk-python/issues/815
```

## Root cause

The warning and the dropped timeout are the same defect. strands' `OllamaModel` / `BedrockModel` constructors absorb any unknown keyword into `**model_config`, where `validate_config_keys()` only emits `warnings.warn(...)` — it **never raises**:

```python
OllamaModel.__init__(host, *, ollama_client_args=None, **model_config)
BedrockModel.__init__(*, boto_client_config=None, ..., **model_config)
```

`model_factory._construct_with_optional_timeout` tried `model_cls(timeout=...)` first and relied on catching `TypeError("unexpected keyword argument")` to fall back to the correct shape. Because `**model_config` swallows the kwarg, that `TypeError` can never fire — so the first attempt always "succeeds", emits the warning, and **drops the timeout before it reaches the HTTP transport**. Only the envelope's secondary wall-clock guard remains, so a genuinely hung HTTP call is no longer cancelled at the transport level.

Verified directly against strands 1.38.0:

| Call | Warning | `model.client_args` (transport) |
|---|---|---|
| `OllamaModel(host, model_id="x", timeout=900)` | yes | `{}` (dropped) |
| `OllamaModel(host, model_id="x", ollama_client_args={"timeout": 900})` | none | `{"timeout": 900}` |

## Correct mechanism

- **Ollama** → `ollama_client_args={"timeout": ...}` (forwarded verbatim to `ollama.AsyncClient(host, **client_args)` → httpx)
- **Bedrock** → `boto_client_config=botocore.Config(read_timeout=..., connect_timeout=...)`

## Scope

Isolated to `backend/agents/investment_team/strategy_lab/agents/model_factory.py`. `llm_service.get_strands_model` is a separate code path (wraps the platform's own client) and is unaffected.

brandonkindred:
Closes #833

## Problem

Strategy Lab logged `Invalid configuration parameters: ['timeout']` on every model construction, and the transport read timeout was **silently dropped** — never reaching the HTTP client.

strands' `OllamaModel` / `BedrockModel` constructors absorb any unknown keyword into `**model_config`, where `validate_config_keys()` only *warns* and never raises. The old `_construct_with_optional_timeout` passed a bare `timeout=` kwarg and relied on catching `TypeError("unexpected keyword argument")` to fall back to the right shape — but that `TypeError` can never fire, so the first attempt always "succeeded", emitted the warning, and dropped the timeout. Only the envelope's secondary wall-clock guard remained, so a hung HTTP call was no longer cancelled at the transport level.

## Fix (`strategy_lab/agents/model_factory.py`)

Replace the broken `TypeError`-probe with `inspect.signature` introspection that only honours an **explicitly-declared** parameter (a name reachable solely via `**kwargs` does not count — strands warns-and-drops it), then route the timeout through each provider's real channel:

- **Ollama** → `ollama_client_args={"timeout": ...}` → `ollama.AsyncClient(host, **client_args)` → httpx
- **Bedrock** → `boto_client_config=botocore.Config(read_timeout=..., connect_timeout=...)`

This removes the warning **and** restores the transport read timeout.

### Review-driven hardening

An adversarial review pass surfaced three secondary issues, all fixed:

- **`int(timeout)` truncation** → botocore now receives the float verbatim (a sub-second timeout no longer collapses to `0`, and the Bedrock path matches the Ollama path).
- **DbC `timeout > 0`** → `_resolve_strands_timeout` now falls back on non-positive env values (honouring its documented positive-float postcondition), and `get_strands_model` asserts `timeout > 0` at the boundary.
- **Misleading docstring** → the `client_args` probe is now honestly documented as a defensive fallback (the client-args name strands' other providers use), not a claim that a current release uses it.

## Verification

- `model_factory.py` at **100%** line coverage; ruff lint + format clean.
- Full `investment_team` suite: **3363 passed, 10 skipped, 0 failures**.
- New regression tests prove the timeout reaches the transport (not merely that construction succeeds), cover both provider channels and their fallbacks, and lock in the precondition + no-truncation behavior.
- Isolated to the Strategy Lab factory; `llm_service.get_strands_model` is a separate code path and unaffected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/backend/agents/investment_team/strategy_lab/agents/model_factory.py
+++ b/backend/agents/investment_team/strategy_lab/agents/model_factory.py
@@ -14,12 +14,18 @@ from __future__ import annotations
 
 import inspect
 import logging
+import math
 import os
 from typing import Any, Dict, Optional
 
 from llm_service.config import resolve_base_url, resolve_model, resolve_provider, resolve_timeout
 
 logger = logging.getLogger(__name__)
+
+# Last-resort transport timeout (seconds) when even ``resolve_timeout`` returns a
+# non-positive / non-finite value. Mirrors the platform default so the resolver's
+# "positive, finite float" postcondition holds unconditionally.
+_DEFAULT_TRANSPORT_TIMEOUT = 900.0
 
 
 def structured_output_enabled() -> bool:
@@ -45,9 +51,13 @@ def _resolve_strands_timeout(agent_key: str) -> float:
     platform-wide ``resolve_timeout`` (which honours ``LLM_TIMEOUT``).
 
     Preconditions: ``agent_key`` is a non-empty model key.
-    Postconditions: returns a positive float — garbage *or non-positive* env
-    values fall back (``float("-5")`` parses cleanly but a non-positive timeout
-    is a misconfiguration, not a valid transport bound).
+    Postconditions: returns a **positive, finite** float, unconditionally.
+    Garbage, non-positive, or non-finite (``inf`` parses cleanly and ``inf > 0``
+    is ``True``, so an infinite read timeout would never cancel a hung call) env
+    values fall back to ``resolve_timeout``; a non-positive/non-finite
+    ``resolve_timeout`` result in turn falls back to
+    ``_DEFAULT_TRANSPORT_TIMEOUT`` so the contract holds even if the platform
+    resolver misbehaves.
     """
     raw = os.environ.get("STRATEGY_LAB_LLM_TIMEOUT")
     if raw is not None and raw.strip() != "":
@@ -56,9 +66,19 @@ def _resolve_strands_timeout(agent_key: str) -> float:
         except ValueError:
             pass
         else:
-            if parsed > 0:
+            if parsed > 0 and math.isfinite(parsed):
                 return parsed
-    return resolve_timeout(agent_key)
+    resolved = resolve_timeout(agent_key)
+    if resolved > 0 and math.isfinite(resolved):
+        return resolved
+    logger.warning(
+        "resolve_timeout(%s) returned a non-positive/non-finite value %r; "
+        "falling back to the default transport timeout %.0fs.",
+        agent_key,
+        resolved,
+        _DEFAULT_TRANSPORT_TIMEOUT,
+    )
+    return _DEFAULT_TRANSPORT_TIMEOUT
 
 
 def _accepts_kwarg(target: Any, name: str) -> bool:
@@ -179,10 +199,11 @@ def get_strands_model(
     prompt-only JSON) and when the toggle is off, so callers can pass a schema
     unconditionally.
 
-    Preconditions: ``agent_key`` is a non-empty model key; ``timeout``, if
-    passed explicitly, is ``> 0`` (a resolved timeout is guaranteed positive by
-    :func:`_resolve_strands_timeout`); ``response_schema``, if given, is a
-    JSON-serializable schema dict.
+    Preconditions: ``agent_key`` is a non-empty model key; ``response_schema``,
+    if given, is a JSON-serializable schema dict. ``timeout``, if passed
+    explicitly, must be a positive, finite number of seconds — a non-positive or
+    non-finite value raises ``ValueError`` (a resolved timeout is guaranteed
+    positive and finite by :func:`_resolve_strands_timeout`).
     Postconditions: returns a constructed strands model. Adding a schema never
     changes which provider/model is selected — only whether the Ollama request
     carries a ``format`` constraint.
@@ -193,9 +214,12 @@ def get_strands_model(
     if timeout is None:
         timeout = _resolve_strands_timeout(agent_key)
     # Boundary enforcement of the construction helpers' ``timeout > 0``
-    # precondition: a non-positive transport timeout is a caller bug (an
-    # explicit bad kwarg), never a value we should forward to httpx/botocore.
-    assert timeout > 0, f"timeout must be > 0 (got {timeout!r})"
+    # precondition: a non-positive or non-finite transport timeout is a caller
+    # bug (an explicit bad kwarg), never a value we should forward to
+    # httpx/botocore. Use an explicit raise rather than ``assert`` so the guard
+    # survives ``python -O``.
+    if not math.isfinite(timeout) or timeout <= 0:
+        raise ValueError(f"timeout must be a positive, finite number of seconds (got {timeout!r})")
 
     use_schema = response_schema is not None and structured_output_enabled()
 
--- a/backend/agents/investment_team/strategy_lab/agents/model_factory.py
+++ b/backend/agents/investment_team/strategy_lab/agents/model_factory.py
@@ -69,11 +69,23 @@ def _resolve_strands_timeout(agent_key: str) -> float:
             if parsed > 0 and math.isfinite(parsed):
                 return parsed
     resolved = resolve_timeout(agent_key)
-    if resolved > 0 and math.isfinite(resolved):
+    # ``resolve_timeout`` is typed ``-> float``, but guard defensively: a
+    # misconfigured resolver returning a non-numeric (str/None) would make
+    # ``resolved > 0`` raise ``TypeError`` and break this function's
+    # unconditional "positive, finite float" postcondition. ``bool`` is excluded
+    # because ``True``/``False`` are not meaningful timeouts (and ``bool`` is an
+    # ``int`` subclass). The ``isinstance`` check also gates ``math.isfinite``,
+    # which itself raises ``TypeError`` on a non-number.
+    if (
+        isinstance(resolved, (int, float))
+        and not isinstance(resolved, bool)
+        and math.isfinite(resolved)
+        and resolved > 0
+    ):
         return resolved
     logger.warning(
-        "resolve_timeout(%s) returned a non-positive/non-finite value %r; "
-        "falling back to the default transport timeout %.0fs.",
+        "resolve_timeout(%s) returned an invalid value %r; falling back to the "
+        "default transport timeout %.0fs.",
         agent_key,
         resolved,
         _DEFAULT_TRANSPORT_TIMEOUT,
--- a/backend/agents/investment_team/strategy_lab/agents/model_factory.py
+++ b/backend/agents/investment_team/strategy_lab/agents/model_factory.py
@@ -12,6 +12,7 @@ are respected consistently with the rest of the platform.
 
 from __future__ import annotations
 
+import inspect
 import logging
 import os
 from typing import Any, Dict, Optional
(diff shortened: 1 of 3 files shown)


backend/agents/investment_team/strategy_lab/agents/model_factory.py
```diff
@@ -124,14 +192,18 @@ def get_strands_model(
     base_url = resolve_base_url()
     if timeout is None:
         timeout = _resolve_strands_timeout(agent_key)
+    # Boundary enforcement of the construction helpers' ``timeout > 0``
+    # precondition: a non-positive transport timeout is a caller bug (an
+    # explicit bad kwarg), never a value we should forward to httpx/botocore.
+    assert timeout > 0, f"timeout must be > 0 (got {timeout!r})"
```
brandonkindred: **[HIGH] standards** — The precondition check `assert timeout > 0` uses an assert statement, which can be disabled with the `-O` flag. In production, this could allow a non-positive timeout to be passed to the transport, potentially causing unexpected behavior (e.g., httpx or botocore may misinterpret 0 or negative values). Input validation should use an explicit `if`/`raise` to guarantee enforcement.

**Suggested fix:** Replace the assert with an explicit check:
```python
if timeout <= 0:
    raise ValueError(f"timeout must be > 0 (got {timeout!r})")
```

backend/agents/investment_team/strategy_lab/agents/model_factory.py
```diff
@@ -44,46 +45,111 @@ def _resolve_strands_timeout(agent_key: str) -> float:
...
     if raw is not None and raw.strip() != "":
         try:
-            return float(raw)
+            parsed = float(raw)
         except ValueError:
             pass
+        else:
+            if parsed > 0:
+                return parsed
     return resolve_timeout(agent_key)
```
brandonkindred: **[MEDIUM] logic** — The function `_resolve_strands_timeout` documents a postcondition that it returns a positive float, but it does not validate the return value of `resolve_timeout`. If `resolve_timeout` returns a non-positive value (e.g., due to misconfiguration), the function will return it, violating its contract. This could lead to a non-positive timeout reaching the transport if the caller does not re-validate (though `get_strands_model` currently asserts, that assert may be disabled).

**Suggested fix:** Either validate the return value of `resolve_timeout` and raise/log a warning if non-positive, or update the docstring to clarify that the postcondition depends on `resolve_timeout`'s contract. For robustness, consider:
```python
resolved = resolve_timeout(agent_key)
if resolved <= 0:
    logger.warning("resolve_timeout returned non-positive value %s; using default", resolved)
    return DEFAULT_TIMEOUT  # or raise
return resolved
```

backend/agents/investment_team/strategy_lab/agents/model_factory.py
```diff
@@ -44,46 +45,111 @@ def _resolve_strands_timeout(agent_key: str) -> float:
...
     """
     raw = os.environ.get("STRATEGY_LAB_LLM_TIMEOUT")
     if raw is not None and raw.strip() != "":
         try:
-            return float(raw)
+            parsed = float(raw)
         except ValueError:
             pass
+        else:
+            if parsed > 0:
```
brandonkindred: **[LOW] logic** — The environment variable parsing accepts any positive float, including `inf`. Passing an infinite timeout to httpx or botocore may cause the transport to never time out, which could lead to hung connections. While unlikely, it's a potential misconfiguration risk.

**Suggested fix:** Add a check for finite values: `if parsed > 0 and math.isfinite(parsed):` or clamp to a reasonable maximum.

brandonkindred (commented):
The PR correctly fixes the timeout forwarding issue by using signature introspection and proper provider-specific channels. The code is well-structured, documented, and tested. However, a few robustness concerns should be addressed: the use of assert for input validation (high), incomplete postcondition enforcement in _resolve_strands_timeout (medium), and lack of infinity guard (low).

**Spec compliance:** The implementation meets the acceptance criteria: the transport timeout is now forwarded correctly, the warning is eliminated, and edge cases (non-positive env values, missing channels) are handled. Tests cover the new behavior and regression scenarios.

backend/agents/investment_team/strategy_lab/agents/model_factory.py
brandonkindred: Fixed in 007cffb6. Replaced the `assert` with an explicit `if not math.isfinite(timeout) or timeout <= 0: raise ValueError(...)` at the boundary, so the guard survives `python -O`. It now also rejects a non-finite explicit `timeout` (inf/nan), not just non-positive. Covered by `test_get_strands_model_rejects_non_positive_or_nonfinite_explicit_timeout` (parametrized over -1/0/inf/nan).

backend/agents/investment_team/strategy_lab/agents/model_factory.py
brandonkindred: Fixed in 007cffb6. `_resolve_strands_timeout` now validates `resolve_timeout`'s return: a non-positive/non-finite result falls back to a new module constant `_DEFAULT_TRANSPORT_TIMEOUT = 900.0` (mirroring the platform default) and logs a warning, so the "positive, finite float" postcondition holds unconditionally rather than depending on the resolver's contract. Covered by `test_resolve_strands_timeout_guards_resolve_timeout_postcondition` (parametrized over 0/-1/inf/nan).

backend/agents/investment_team/strategy_lab/agents/model_factory.py
brandonkindred: Fixed in 007cffb6. The env parse now requires `parsed > 0 and math.isfinite(parsed)`, so `inf`/`-inf`/`nan` fall back to `resolve_timeout` instead of forwarding an infinite read timeout that would never cancel a hung call. Covered by `test_resolve_strands_timeout_non_positive_or_nonfinite_value_falls_back` (parametrized to include inf/-inf/nan).

backend/agents/investment_team/strategy_lab/agents/model_factory.py
```diff
@@ -44,46 +51,125 @@ def _resolve_strands_timeout(agent_key: str) -> float:
...
-            return float(raw)
+            parsed = float(raw)
         except ValueError:
             pass
-    return resolve_timeout(agent_key)
+        else:
+            if parsed > 0 and math.isfinite(parsed):
+                return parsed
+    resolved = resolve_timeout(agent_key)
+    if resolved > 0 and math.isfinite(resolved):
```
brandonkindred: **[MEDIUM] logic** — The function `_resolve_strands_timeout` documents a postcondition of returning a positive finite float unconditionally, but if `resolve_timeout` returns a non-numeric value (e.g., a string due to a misconfiguration), the comparison `resolved > 0` will raise a `TypeError`, violating the postcondition and potentially crashing the calling code. The function should handle non-numeric returns gracefully by falling back to the default timeout.

**Suggested fix:** Wrap the comparison in a try-except block or check `isinstance(resolved, (int, float))` before comparing. For example:
```python
try:
    if resolved > 0 and math.isfinite(resolved):
        return resolved
except TypeError:
    pass
```
or
```python
if isinstance(resolved, (int, float)) and resolved > 0 and math.isfinite(resolved):
    return resolved
```

backend/agents/investment_team/tests/test_misc_helpers.py
```diff
@@ -681,17 +682,24 @@ def test_get_strands_model_bedrock_branch(monkeypatch: pytest.MonkeyPatch) -> No
...
 
     import strands.models as strands_models
 
     monkeypatch.setattr(strands_models, "BedrockModel", _StubBedrock)
     result = model_factory.get_strands_model()
     assert isinstance(result, _StubBedrock)
-    assert result.kwargs["model_id"] == "anthropic.claude-3-haiku-20240307-v1:0"
+    assert result.model_id == "anthropic.claude-3-haiku-20240307-v1:0"
+    assert result.boto_client_config is not None
+    assert result.boto_client_config.read_timeout == 77
```
brandonkindred: **[INFO] testing** — The assertion `assert result.boto_client_config.read_timeout == 77` compares an int to a float timeout. While functionally equivalent, it would be clearer to assert `== 77.0` to explicitly match the float type used throughout the timeout handling.

**Suggested fix:** Change the assertion to `assert result.boto_client_config.read_timeout == 77.0` for consistency.

brandonkindred (commented):
The code is well-structured, thoroughly documented, and includes comprehensive tests. The fix correctly addresses the timeout forwarding issue by using signature introspection to route the timeout through the appropriate client-args channels. The hardening measures (float preservation, precondition enforcement) are solid. A minor robustness gap exists in `_resolve_strands_timeout` where a non-numeric return from `resolve_timeout` could cause an unexpected TypeError; otherwise, the implementation is production-ready.

**Spec compliance:** The implementation meets the acceptance criteria: the transport timeout is now forwarded to the HTTP client (Ollama via ollama_client_args, Bedrock via boto_client_config), the invalid configuration warning is eliminated, and the timeout is correctly applied. All specified hardening (float timeout, DbC enforcement, docstring correction) is included. The full test suite passes with 100% line coverage on the changed file.

backend/agents/investment_team/strategy_lab/agents/model_factory.py
brandonkindred: Fixed in 3f17ce83. `_resolve_strands_timeout` now validates the `resolve_timeout` result with `isinstance(resolved, (int, float)) and not isinstance(resolved, bool)` **before** the `math.isfinite`/`> 0` checks, so a non-numeric (str/None) return falls back to `_DEFAULT_TRANSPORT_TIMEOUT` (with a warning) instead of raising `TypeError`. The unconditional "positive, finite float" postcondition now holds for any resolver return. `test_resolve_strands_timeout_guards_resolve_timeout_postcondition` is parametrized over `"not-a-number"`, `None`, and `True` in addition to `0`/`-1`/`inf`/`nan`.

backend/agents/investment_team/tests/test_misc_helpers.py
brandonkindred: Fixed in 3f17ce83. Changed the assertion to `== 77.0` to explicitly match the float type forwarded throughout the timeout handling.

backend/agents/investment_team/strategy_lab/agents/model_factory.py
```diff
@@ -44,46 +51,137 @@ def _resolve_strands_timeout(agent_key: str) -> float:
...
         try:
-            return float(raw)
+            parsed = float(raw)
         except ValueError:
             pass
-    return resolve_timeout(agent_key)
+        else:
+            if parsed > 0 and math.isfinite(parsed):
+                return parsed
+    resolved = resolve_timeout(agent_key)
```
brandonkindred, line 77: **[MEDIUM] logic** — The function `_resolve_strands_timeout` documents an unconditional postcondition of returning a positive, finite float. However, it calls `resolve_timeout(agent_key)` without a try-except block. If `resolve_timeout` raises an exception (e.g., due to a misconfiguration or bug), the exception propagates, violating the unconditional postcondition and potentially crashing the model construction. The function should catch exceptions from `resolve_timeout` and fall back to `_DEFAULT_TRANSPORT_TIMEOUT` to truly guarantee the postcondition.

**Suggested fix:** Wrap the `resolve_timeout` call in a try-except block. For example:
```python
try:
    resolved = resolve_timeout(agent_key)
except Exception as exc:
    logger.warning(
        "resolve_timeout(%s) raised %r; falling back to default transport timeout %.0fs.",
        agent_key, exc, _DEFAULT_TRANSPORT_TIMEOUT,
    )
    return _DEFAULT_TRANSPORT_TIMEOUT
```
Then proceed with the existing validation of `resolved`.

backend/agents/investment_team/strategy_lab/agents/model_factory.py
```diff
@@ -124,14 +225,21 @@ def get_strands_model(
     base_url = resolve_base_url()
     if timeout is None:
         timeout = _resolve_strands_timeout(agent_key)
+    # Boundary enforcement of the construction helpers' ``timeout > 0``
+    # precondition: a non-positive or non-finite transport timeout is a caller
+    # bug (an explicit bad kwarg), never a value we should forward to
+    # httpx/botocore. Use an explicit raise rather than ``assert`` so the guard
+    # survives ``python -O``.
+    if not math.isfinite(timeout) or timeout <= 0:
```
brandonkindred, line 243: **[LOW] logic** — The `get_strands_model` function's boundary check uses `math.isfinite(timeout)`. If `timeout` is not a number (e.g., a string), `math.isfinite` raises `TypeError` instead of the documented `ValueError`. The docstring states that a non-positive or non-finite value raises `ValueError`, but does not account for non-numeric types. While the function is internal and callers are expected to pass numbers, the contract is incomplete. Consider adding an explicit type check or documenting the `TypeError` possibility.

**Suggested fix:** Either add a type check before the `math.isfinite` call, e.g.:
```python
if not isinstance(timeout, (int, float)) or isinstance(timeout, bool):
    raise TypeError(f"timeout must be a number, got {type(timeout).__name__}")
if not math.isfinite(timeout) or timeout <= 0:
    raise ValueError(...)
```
or update the docstring to mention that non-numeric values will raise `TypeError`.

brandonkindred (commented):
The PR effectively fixes the transport timeout forwarding bug by using signature introspection and proper channels. The code is well-structured, documented, and tested. However, there is a medium issue with the unconditional postcondition of `_resolve_strands_timeout` not being fully enforced due to missing exception handling. A low issue exists with the boundary check's contract. Overall, the implementation is solid and meets the acceptance criteria.

**Spec compliance:** The code implements the required fix: timeout is now forwarded via `ollama_client_args` for Ollama and `boto_client_config` for Bedrock, eliminating the warning and dropped timeout. The acceptance criteria are met. The additional hardening (DbC, no truncation) is properly implemented. The tests cover the new functionality.

backend/agents/investment_team/strategy_lab/agents/model_factory.py
brandonkindred: Thanks — I looked at `resolve_timeout` (`llm_service/config.py:201`) and it is a **total function**: it reads `os.environ.get(ENV_LLM_TIMEOUT) or "900"` and its only failure mode (`float()` on a bad string) is caught internally, returning `900.0`. There is no code path on which it raises, so a `try/except` here would be guarding an impossible exception.

It would also conflict with the project DbC rule (CLAUDE.md): *"never try/except around a contract failure to hide it."* `resolve_timeout` is typed `-> float`; if a future change made it violate that contract by raising, that bug should surface loudly rather than be silently swallowed into a 900s default (which could mask a real misconfiguration). The existing value-guard (`isinstance` + `math.isfinite` + `> 0`, added previously) already covers every value it can return.

I have not added the try/except, but to make the decision explicit for future readers/reviewers I documented the totality reasoning in a code comment right above the call (4ac7de1d). Happy to revisit if `resolve_timeout` ever gains a raising path.

backend/agents/investment_team/strategy_lab/agents/model_factory.py
brandonkindred: Fixed in 4ac7de1d. Added an explicit type check before `math.isfinite`: `if not isinstance(timeout, (int, float)) or isinstance(timeout, bool): raise TypeError("timeout must be a number of seconds (got <type>)")`. A non-numeric kwarg now raises a clear `TypeError` instead of an obscure one from `math.isfinite`, and the docstring precondition documents that non-numeric -> `TypeError` while non-positive/non-finite -> `ValueError`. Covered by `test_get_strands_model_rejects_non_numeric_explicit_timeout` (parametrized over str/bool/list).

brandonkindred (commented):
The code correctly fixes the timeout forwarding bug by using inspect.signature to detect explicit parameters and routing the timeout through the proper channels (ollama_client_args for Ollama, boto_client_config for Bedrock). The implementation follows Design by Contract with clear preconditions/postconditions, is well-documented, and has comprehensive test coverage (100% line coverage). No security, performance, or integration issues found. The code is production-ready.

**Spec compliance:** The code fully meets the acceptance criteria: the transport timeout is no longer dropped, the invalid configuration warning is eliminated, and the timeout is correctly forwarded to httpx/botocore. All specified provider channels and fallbacks are implemented. The regression tests prove the timeout reaches the transport.

PR #839, opened 2026-06-10
LLM client: proof-of-change thinking-downgrade retry for empty responses

### LINKED ISSUE #774: LLM client: empty-content / reasoning-only responses retry at max thinking, burning the full retry budget
## Problem

When a model spends its whole token budget on reasoning and returns **zero content**, the Ollama client treats it as a transient error and retries — but each retry re-sends the request at the *same* (often maximum) thinking level, so the degenerate case repeats up to `LLM_MAX_RETRIES` (default 10) times, each attempt burning the full budget again before failing.

- The empty-content path raises `LLMTemporaryError` (`backend/agents/llm_service/clients/ollama.py:664-670`): *"Empty response from LLM; treating as transient for retry."*
- A sibling warning already diagnoses the root cause (`ollama.py:827-832`): *"LLM produced reasoning tokens but no content … model likely spent its token budget on thinking. Consider raising max_tokens or passing think=False."*
- The retry loop re-issues the identical payload with backoff (`ollama.py` ~951-964); there is no thinking reduction between attempts, so a reasoning-only response is retried at max effort every time.

This is the expensive tail of the coding-team cost work: a single stuck call can consume `LLM_MAX_RETRIES + 1` full-budget generations with no change in inputs.

## Proposed fix

- When a call finishes with **reasoning-only / zero content**, retry **once with reduced thinking** (e.g. `think=False`, or one level down for models with registered thinking levels) instead of repeating at max effort for the whole retry count. A response that has content but is otherwise transient keeps the existing retry behaviour.
- Apply in the shared client so every team benefits (this is the deferred, shared-surface half of the coding-team cost guardrails work, split out so the retry-path change can be reviewed on its own).
- Keep it observable: log the downgrade so operators can see when it fires.

## Acceptance criteria

- A reasoning-only / empty-content response triggers at most one reduced-thinking retry; the original max-thinking payload is not re-sent `LLM_MAX_RETRIES` times (regression test with a stub transport asserting the `think`/`reasoning_effort` of each attempt).
- Genuine transient errors (5xx, connection reset, timeout) keep the existing exponential-backoff retry behaviour unchanged.
- The reduced-thinking retry is logged.
- 90% line coverage on new/modified code.

brandonkindred:
Closes #774

## Problem

A 200 response with zero assistant content (typically a reasoning-only response from a thinking model) raised `LLMTemporaryError` and the retry loop re-sent the **identical** payload — same `think`/`reasoning_effort` — up to `LLM_MAX_RETRIES` (default 10) times. Nothing changed between attempts, so the degenerate outcome repeated every time. As discussed in the issue comments, this is a **proof-of-change** problem, not a transport-retry problem.

## Change

Empty 200 responses are now a third failure class — **semantic exhaustion** — with its own budget, separate from the transient (5xx/connection/timeout) and 429 schedules:

1. **One immediate proof-of-change retry with reduced thinking** (no backoff sleep — the changed payload is the proof):
   - models with registered thinking levels: one level down (e.g. `max` → `high`)
   - boolean `think=True` / unregistered level strings: `think=False` (`reasoning_effort="none"`)
2. **Fail hard after the downgrade** (or immediately when thinking is already off / at the lowest level, leaving no provable change): a new `LLMSemanticExhaustionError` carries a structured receipt — `failure_class="semantic_exhaustion"`, `attempts_used`, `original_thinking_level`, `retry_thinking_level`, `content_bytes_seen`, `finish_reason`, and a last-payload fingerprint. The receipt is logged at ERROR; the downgrade retry is logged at WARNING; all three entry points (`complete_json`, `complete`, `chat`) record `error_type=semantic_exhaustion` telemetry.
3. **Transport faults and 429s keep their existing independent retry schedules**, before and after the downgrade (a 5xx after the downgrade retries the downgraded payload). The truncated-with-partial-content continuation path is preserved — and continuations now **resume at the thinking level that actually produced the partial content** (`LLMTruncatedError.think_used`), so a call downgraded mid-flight is not silently continued at the original, just-failed level.

`LLMSemanticExhaustionError` subclasses `LLMTemporaryError` so existing pause/degrade handlers keep working, and is exported from the `llm_service` package namespace alongside its siblings. The no-loop guarantee is structural inside the client (the raise happens in the dedicated `except` handler, which a sibling `except LLMTemporaryError` can never catch) **and** honored by the Strategy Lab fault-tolerance envelope, which now classifies semantic exhaustion as fatal instead of macro-retrying the identical prompt with a fresh downgrade budget.

**Kill switch:** `LLM_THINKING_DOWNGRADE_RETRY` (default on) — a falsy value restores the legacy transient-schedule behavior verbatim. The policy decision (downgrade vs legacy) lives solely in `_ollama_post`'s retry handler; `_parse_response_content` only signals.

## Files

- `backend/agents/llm_service/interface.py` — `LLMSemanticExhaustionError` with the receipt fields; `LLMTruncatedError.think_used`
- `backend/agents/llm_service/config.py` — pure `downgrade_think(model, think)` helper, shared `env_flag_enabled()` bool-toggle parser, env constant
- `backend/agents/llm_service/clients/ollama.py` — internal `_EmptyResponseSignal`, downgrade/legacy policy in `_ollama_post`, `resolved_think` threading, telemetry parity, think-level propagation into continuations
- `backend/agents/llm_service/util.py` + `structured.py` — shared `sha256_fingerprint()` helper backing the payload fingerprint and the structured-output prompt hash
- `backend/agents/llm_service/__init__.py` — public export
- `backend/agents/investment_team/strategy_lab/agents/_llm_envelope.py` — semantic exhaustion classified as fatal (non-retriable)
- Tests: regression tests asserting the per-attempt `think`/`reasoning_effort` of every payload (stubbed transport), fail-hard receipt fields, transient-schedule interplay, kill switch, continuation think-level inheritance, telemetry parity, envelope classification, and table-driven `downgrade_think`/`env_flag_enabled` units
- Docs: `docs/ENV_VARS.md` + `CLAUDE.md` env index

## Testing

- `ruff check` passes; modified files formatted.
- `pytest agents/llm_service/tests/` — 199 passed; `pytest agents/investment_team/tests/test_strategy_lab_llm_envelope.py` — 57 passed. Two pre-existing failures in this environment are unrelated and fail on a clean checkout too: `test_extract_json_json_repair_unescaped_quotes_in_strings` (missing optional json-repair dependency locally) and `test_strands_provider.py` collection (imports `_resolve_ollama_auth_headers`, which does not exist in `strands_provider.py`).
- All new/modified lines are covered (uncovered ranges in the coverage report are pre-existing paths).

https://claude.ai/code/session_01Lv4gGxTHeBvaEDhLzxGQDX


backend/agents/llm_service/clients/ollama.py
```diff
@@ -1139,12 +1280,7 @@ def _ollama_post(
                     raise LLMRateLimitError(str(e), status_code=429, cause=e)
```
brandonkindred, line 1284: **[CRITICAL] logic** — The 429 rate-limit retry logic is broken. When a 429 response is received, the code in the `except httpx.HTTPStatusError` block (line 1280) raises `LLMRateLimitError`. However, this exception is raised inside an except block, so it is not caught by the sibling `except LLMRateLimitError` handler (line 1156). As a result, the rate-limit backoff and retry never execute; the `LLMRateLimitError` propagates immediately to the caller, bypassing the configured `LLM_RATE_LIMIT_MAX_RETRIES` and backoff schedule. This defeats the independent rate-limit retry budget and will cause any 429 response to fail the call without retry.

**Suggested fix:** Move the 429 check and `LLMRateLimitError` raise into the `try` block before the generic `httpx.HTTPStatusError` raise. For example, after reading the response body, check `if response.status_code == 429: raise LLMRateLimitError(...)` instead of raising `HTTPStatusError`. This way the `except LLMRateLimitError` handler will catch it and perform the backoff. Also ensure that `rl_log_body` and `rl_log_headers` are set before raising (or attach them to the exception) so the handler can log them. Remove the 429 handling from the `except httpx.HTTPStatusError` block to avoid dead code.

backend/agents/llm_service/clients/ollama.py
```diff
@@ -1067,7 +1144,14 @@ def _ollama_post(
...
+            except LLMPermanentError:
+                raise
+            except LLMTruncatedError as e:
+                # Stamp the thinking level of the attempt that produced the
+                # partial content so continuation requests resume at the SAME
+                # level — a call downgraded mid-flight must not be silently
+                # continued at the original (just-failed) thinking level.
+                e.think_used = active_think
                 raise
             except LLMRateLimitError as e:
```
brandonkindred, line 1160: **[HIGH] logic** — The `except LLMRateLimitError` handler references `rl_log_body` and `rl_log_headers` (line 1173) which are only set in the `except httpx.HTTPStatusError` block for 429. If the handler is ever reached from a different code path (e.g., if `LLMRateLimitError` is raised elsewhere), these variables will be unbound, causing a `NameError` at runtime. Even after fixing the 429 flow, these variables should be properly scoped or passed via the exception to avoid potential crashes.

**Suggested fix:** Either initialize `rl_log_body` and `rl_log_headers` to `None` before the loop, or attach the response body/headers to the `LLMRateLimitError` instance and access them from the exception in the handler. This ensures the handler is robust regardless of how the exception is raised.

backend/agents/llm_service/clients/ollama.py
```diff
@@ -772,31 +805,77 @@ def _ollama_post(
...
+        # always terminates by returning or raising.
         transient_attempt = 0
         rate_limit_attempt = 0
-        max_total_attempts = max_retries + rate_limit_max_retries + 1
+        # Semantic-exhaustion state: `semantic_attempt > 0` means the one
+        # proof-of-change downgrade retry has been spent for this call.
+        semantic_attempt = 0
+        active_think: "bool | str | None" = resolved_think
+        any_content_bytes = False
+        max_total_attempts = max_retries + rate_limit_max_retries + 2
```
brandonkindred: **[LOW] logic** — The `max_total_attempts` variable is calculated as `max_retries + rate_limit_max_retries + 2`, but it does not account for the semantic retry attempt. While only used for logging, it may display misleading attempt counts (e.g., "attempt 3 of 17" when the actual maximum could be higher if a semantic retry occurs). This is minor but could cause confusion during debugging.

**Suggested fix:** Consider computing `max_total_attempts` as `max_retries + rate_limit_max_retries + (2 if _thinking_downgrade_enabled() else 1)` or simply not including a total in the log messages, or using a dynamic total. Alternatively, document that the total is approximate.

brandonkindred (commented):
The PR introduces a well-designed proof-of-change retry for empty LLM responses, with thorough documentation and tests. However, a critical bug in the refactored exception handling breaks the existing 429 rate-limit retry logic, causing rate-limited requests to fail immediately without retries. Additionally, the rate-limit handler references potentially uninitialized variables. These issues must be fixed before merging.

**Spec compliance:** The implementation meets the specified requirements for semantic exhaustion handling, including the downgrade retry, kill switch, telemetry, and continuation think-level propagation. The integration with the strategy lab envelope and orchestrator is correct. However, the independent rate-limit retry schedule is broken, which violates the requirement that "Transport faults and 429s keep their existing independent retry schedules".

- `backend/agents/llm_service/tests/test_ollama_client.py` — **[MEDIUM] testing** — The existing rate-limit retry test (likely `test_ollama_rate_limit_retries_with_backoff`) may not be exercising the actual production code path due to the bug, or it may be passing because the mock setup bypasses the `httpx.HTTPStatusError` flow. If the test is passing despite the bug, it indicates a gap in test coverage: the test does not accurately simulate a real 429 response flow. This could allow the critical bug to go unnoticed.

**Suggested fix:** Update the rate-limit tests to use a mock that returns a response with status 429 and triggers the `httpx.HTTPStatusError` path (or the new direct `LLMRateLimitError` raise). Ensure the test verifies that the configured number of retries and backoff occur, and that the final error is raised after exhaustion. This will catch any regression in the rate-limit handling.

backend/agents/llm_service/clients/ollama.py
brandonkindred: I believe this one is a false positive — the backoff does execute, on both 429 paths, and the existing test suite pins it:

1. **The primary 429 path raises from the `try` body, not from an `except` handler.** The `if status == 429:` branch (~line 1065) sits in the status-handling code inside the `try`, so its `raise LLMRateLimitError(...)` *is* caught by the loop-level `except LLMRateLimitError` handler, which owns the backoff sleep. (The "sibling except can't catch" rule only applies to raises *inside* an except block.) This is deliberate: raising first unwinds the `with sem / with Client / with stream` contexts so the multi-minute 429 sleep never holds the concurrency slot or the open stream. Pinned by `test_ollama_429_then_success`, `test_ollama_429_exhaustion_raises`, and `test_ollama_429_sleep_runs_after_stream_released`.

2. **The `except httpx.HTTPStatusError` 429 branch at line 1280 doesn't skip the schedule either** — it applies the rate-limit backoff *inline* (`_rate_limit_backoff_sleep` → `rate_limit_attempt += 1` → `continue`, using the same shared counter) and only raises `LLMRateLimitError` once the budget is exhausted, which is the intended terminal raise. The inline handling exists precisely because a raise from this handler couldn't reach the sibling clause — that's what the in-code comment above it explains. Pinned by `test_ollama_httpstatuserror_429_uses_rate_limit_schedule` and `test_ollama_httpstatuserror_429_exhaustion_raises`.

The suggested fix (raise `LLMRateLimitError` from the try body) is what path 1 already does; it can't apply to path 2 because `httpx.HTTPStatusError` is raised by httpx itself mid-`try`, so the except handler is the only place to classify it. Happy to add a comment cross-referencing the two paths if that would make the structure easier to scan.

---

backend/agents/llm_service/clients/ollama.py
brandonkindred: This can't NameError — both variables are initialized before the loop and the handler guards for the unset case:

- Line 846-847 (before `while True`): `rl_log_body: Optional[str] = None` / `rl_log_headers: Any = None`
- They're populated in the **in-try** 429 branch (~line 1072, `rl_log_body = response.text`), not in the `except httpx.HTTPStatusError` block — that handler logs via `_log_llm_server_error` with the response it already holds and never touches these.
- The `except LLMRateLimitError` handler reads them behind `if rl_log_body is not None:` (line 1171), so an `LLMRateLimitError` raised from any other path (e.g. without a captured body) just skips the exhaustion log.

So every path is bound and guarded. Attaching the body/headers to the exception would be a reasonable alternative shape, but it would carry response payloads on an exception that outlives the response for no behavioral gain — the current None-init + guard covers the same cases.

---

backend/agents/llm_service/clients/ollama.py
brandonkindred: Good catch on the kill-switch case. One correction to the premise: the `+2` *does* include the semantic retry slot (it's `1` first attempt + `1` downgrade retry on top of the two budgets — it was bumped from `+1` for exactly that). The real imprecision was the opposite direction: with `LLM_THINKING_DOWNGRADE_RETRY` disabled, no downgrade slot exists and the denominator overstated by one.

Fixed in 6806584 with your suggested shape: `max_retries + rate_limit_max_retries + (2 if _thinking_downgrade_enabled() else 1)` — exact in both modes now. (Log-only either way; the receipt's `attempts_used` counts actual HTTP attempts via the summed per-budget counters and was already correct in mixed sequences.)

---

brandonkindred (commented):
The PR implements the proof-of-change thinking-downgrade retry for empty LLM responses effectively, with thorough testing and documentation. The code is well-structured and follows Design by Contract. However, there are some maintainability concerns: the HTTP error handling in `_ollama_post` contains duplicated and potentially dead code paths for 5xx and 429 errors, which could confuse future developers and lead to bugs if the active path is changed. Additionally, the `LLMSemanticExhaustionError` inheritance from `LLMTemporaryError` while being treated as non-retriable is a minor design smell. Overall, the implementation meets the acceptance criteria and is production-ready, but the error handling duplication should be cleaned up.

**Spec compliance:** The code fully implements the specified behavior: empty 200 responses trigger a single proof-of-change retry with reduced thinking, then fail with `LLMSemanticExhaustionError`; the kill switch restores legacy behavior; transport and rate-limit retries remain independent; continuation requests resume at the downgraded thinking level; telemetry is recorded for all entry points; and the envelope classifies semantic exhaustion as fatal. All acceptance criteria are satisfied.

backend/agents/llm_service/clients/ollama.py
```diff
@@ -1089,19 +1173,76 @@ def _ollama_post(
...
+                    content_bytes_seen=any_content_bytes,
+                    payload_fingerprint=fingerprint,
+                    finish_reason=sig.finish_reason,
+                )
+            except LLMTemporaryError as e:
+                last_error = e
+                if _retry_transient_step(str(e)):
                     continue
                 raise last_error
             except httpx.HTTPStatusError as e:
```
brandonkindred, line 1248: **[MEDIUM] standards** — The retry loop in `_ollama_post` contains two separate mechanisms for handling HTTP error status codes: manual checks in the `try` block (e.g., raising `LLMTemporaryError` for 5xx around line 1095) and an `except httpx.HTTPStatusError` block (line 1248) that also handles 5xx and 429. This duplication makes the error handling inconsistent and harder to maintain. The 429 case in the `except` block (line 1276) raises `LLMRateLimitError` directly, which would bypass the dedicated rate-limit retry logic in `except LLMRateLimitError` (line 1156) if that path were ever triggered. Currently, one of these paths is likely dead code, but the duplication risks future bugs if the active error-handling strategy is changed without updating both places.

**Suggested fix:** Consolidate HTTP error handling into a single strategy. Either handle all status codes in the `try` block by checking `response.status_code` and raising the appropriate `LLM*` exceptions (so they are caught by the dedicated `except` handlers), or use `response.raise_for_status()` and handle everything in `except httpx.HTTPStatusError`. Remove the unused path to eliminate dead code and prevent confusion. Ensure that 429 errors are always processed by the rate-limit retry schedule.

backend/agents/llm_service/interface.py
```diff
@@ -61,6 +61,61 @@ class LLMUnreachableAfterRetriesError(LLMTemporaryError):
     """Raised when the caller exhausted retries and could not reach the LLM. Orchestrator should pause job."""
 
 
+class LLMSemanticExhaustionError(LLMTemporaryError):
```
brandonkindred, line 64: **[LOW] standards** — `LLMSemanticExhaustionError` subclasses `LLMTemporaryError` for compatibility with existing pause/degrade handlers, but it is treated as a terminal (non-retriable) error in the envelope and `call_llm_with_retries`. This violates the Liskov substitution principle because a subclass is not substitutable for its parent in all contexts. While intentional, it may surprise future maintainers who expect all `LLMTemporaryError` subclasses to be retriable.

**Suggested fix:** Add a prominent note in the class docstring explaining that despite inheriting from `LLMTemporaryError`, this error is considered fatal and should not be retried at the macro level. Alternatively, consider refactoring the exception hierarchy to avoid this violation, but that may be a larger change.

backend/agents/llm_service/clients/ollama.py
```diff
@@ -761,6 +792,8 @@ def _ollama_post(
         rate_limit_initial: float,
```
brandonkindred, line 792: **[INFO] standards** — The `_ollama_post` method is very long (over 500 lines) and contains complex retry logic with three independent budgets. This reduces readability and makes unit testing individual retry policies harder.

**Suggested fix:** Consider extracting the retry logic for each failure class (transient, rate limit, semantic exhaustion) into separate helper methods or a retry manager class to improve modularity and testability.

backend/agents/llm_service/clients/ollama.py
brandonkindred, line 845: **[INFO] standards** — The variable `max_total_attempts` is computed as `max_retries + rate_limit_max_retries + 2` but is only used in a log message (line 869). The loop does not enforce this bound; it relies on the individual budgets to terminate. This could be misleading.

**Suggested fix:** Add a comment clarifying that `max_total_attempts` is for logging purposes only and that the loop termination is guaranteed by the individual retry budgets. Alternatively, enforce the bound as a safety net.

PR #841, opened 2026-06-10
Cognition Step 10: invoke gate — idempotent inject-on-invoke / writeback-on-return at the boundary

### LINKED ISSUE #743: [Cognition][Step 10] Invoke proxy integration
**Milestone:** C — Tools & orchestration (#726)
**Depends on:** #741 (Step 8), #732 (Step 9)
**Plan:** `IMPLEMENTATION_PLAN.md` (PR #723), Step 10.

**Goal:** Make the inject-on-invoke / writeback-on-return contract live at the boundary.

**Files:** `unified_api/routes/agents.py` — lazy catch-up → load context → advisory-into-system-prompt + enforced precondition gate (block → 4xx + memory event) → inject `cognition` block → on return persist writeback + postcondition check. Helper for in-process teams (no HTTP hop).

**✅ Acceptance:** stubbed-sandbox test shows inject + persist; precondition block → 4xx; postcondition path covered; ruff clean; ≥90% coverage.

brandonkindred:
Closes #743

Makes the cognition **inject-on-invoke / writeback-on-return contract live at the invoke boundary** (`IMPLEMENTATION_PLAN.md` Step 10). The first commit is the implementation plan (`backend/agents/plans/cognition_step10_invoke_proxy_integration.plan.md`); the second is the implementation; the third applies the code-review findings.

## What's new

**`agent_cognition/invoke_gate.py`** — the reusable lifecycle orchestrator over the Step 8 `CognitiveContext` facade, shared by the HTTP proxy and a new `invoke_in_process(agent_id, body, runner)` helper for in-process teams (no HTTP hop, identical contract). Pure orchestration: typed `GateOutcome`s, no FastAPI types. 100% line coverage.

**Run-once idempotency**
- `source_run_id` = caller `Idempotency-Key` header, else SHA-256 of the body's canonical JSON (byte-identical keyless retries still dedup).
- `claim_run` against the leased `agent_cognition_runs` ledger: a terminal row with a matching hash **replays the stored envelope without re-invoking** — including `blocked` rows, so a retried rule block replays the same 4xx (replays carry `X-Khala-Replayed: true`, skip the sandbox post and don't write duplicate console run rows); hash mismatch or still-valid lease → **409**; expired lease reclaims in place and re-executes.
- `requires_idempotency_key: true` manifests reject keyless invokes **400**; on a ledger outage they fail closed **503**, while other cognition agents degrade to an unledgered invoke (a cognition outage never breaks the invoke).

**Gates and persistence ordering** (`unified_api/routes/agents.py` reworked into a thin adapter)
- Lazy rollup catch-up → context load via the facade → enforced **precondition** (a block persists an `error` memory event + the blocked ledger envelope) → marker-envelope wrap → **full-envelope re-check** against `AGENT_INVOKE_MAX_PAYLOAD_BYTES` (413 — the side channel shares the payload budget).
- On return the enforced **postcondition runs before any persistence** — fixing the prior ordering bug where the console run row stored the model output before the gate evaluated it. A violation drops the output and agent-authored memory everywhere but persists the shim's trusted `tool_audit` + a blocked-run event. A pass persists defensively-validated `memory_events` and completes the ledger with `{status_code, content}` (stored without the per-invoke `sandbox` block so a replay can't masquerade as a fresh boot).
- The **warming 202 path deliberately precedes the gate** — claiming a leased run and then answering "retry shortly" would make every warm-up poll 409 against its own claim (regression-tested).
- Non-cognition agents keep today's pass-through behavior exactly.

## Code-review hardening (third commit)

- **Attempt-scoped memory keys** — events persist under `{source_run_id}#{attempt_token}`, so an expired-lease re-execution's *new* side-effect events can never silently dedup against a failed attempt's rows on the `(agent_id, run_id, seq)` `ON CONFLICT` key.
- **Structured Pydantic outputs in `invoke_in_process`** — a runner returning a `BaseModel` (top-level or nested) is dumped to a dict, not stringified, so enforced postconditions still apply and the replay envelope stays structured.
- **Claim release** — new `context.abandon_run` (claim-token-fenced DELETE) frees the claim where the agent provably produced nothing to protect (oversize wrapped envelope 413; over-cap upstream response 502): immediate retries re-execute instead of 409-ing until lease expiry. Transport errors deliberately **hold** the lease as the concurrent double-run guard.
- **Budgeted lazy catch-up** — `ensure_rollups_current` gains `max_periods` (+ `RollupReport.truncated`); the gate passes `AGENT_COGNITION_INVOKE_ROLLUP_BUDGET` (default 8, each period ≈ one LLM call), so a cold-start backlog can't run hundreds of sequential LLM calls inline on the hot path.
- **Fail-closed postconditions** — a 2xx response without an object `output` on an agent carrying enforced postcondition rules now blocks (rules evaluate against `{}`) instead of silently skipping the gate.
- **Single status map** (`GATE_ERROR_HTTP_STATUS`) shared by route + in-process helper; blocked outcomes carry structured `block_phase`/`block_reason` so the route never parses the stored HTTP body; shared `FakeRunLedger` test double in `agent_cognition/testing.py`; dead field + duplicate branch removed.

**Scoping note:** proxy-side `platform_bound` tool-loop driving stays gated on the generated-agent runtime scaffold (Step 14) — `drive_platform_bound_loop` already landed with the tools layer.

## Tests & quality
- `agent_cognition/tests/test_invoke_gate.py` — 47 cases, **100% line coverage** on the gate and the shared test double; `test_context.py` gains abandon-run fencing (live-PG) tests; `test_rollup.py` gains the budget-cap test; the route acceptance matrix covers replay / 409 / 400 / 413 / 422 / warming-202-without-claim / outage degradation / oversize-claim-release / non-cognition bypass.
- `ruff check` + `ruff format` clean; the only failures in broader local runs are pre-existing on `main` (`test_dispatch.py`, env-dependent `unified_api` tests — fail identically on a clean tree).
- Docs: `agent_cognition/README.md` ("Invoke gate": idempotency table, claim-release semantics, attempt-scoped keys, fail-closed postconditions, rollup budget), `docs/ENV_VARS.md` + `CLAUDE.md` (new `AGENT_COGNITION_INVOKE_ROLLUP_BUDGET`), CHANGELOG.

https://claude.ai/code/session_01JECoSxdcQghattEBo1jLs5


brandonkindred (commented):
The PR implements the cognition invoke gate (Step 10) with high quality. The code is well-architected, follows Design by Contract, has thorough documentation and tests (100% line coverage on the gate). No security, performance, or integration issues found. The implementation meets all acceptance criteria and integrates cleanly with the existing route.

**Spec compliance:** All acceptance criteria from the implementation plan are met: idempotency with Idempotency-Key, run ledger claim/replay/conflict/reclaim, lazy rollup catch-up with budget, precondition gate with durable blocked envelope, full-envelope payload re-cap, postcondition-before-persist ordering, writeback persistence with attempt-scoped keys, in-process helper, and proper degradation on storage outage. The route is reworked as a thin adapter. Tests cover the full acceptance matrix.

backend/agents/agent_cognition/invoke_gate.py
```diff
@@ -0,0 +1,1009 @@
...
+        try:
+            claim = await asyncio.to_thread(
+                claim_run, agent_id, source_run_id, request_hash, default_run_lease()
+            )
+        except AgentCognitionStorageUnavailable as exc:
+            outage = _storage_outage_outcome(agent_id, requires_idempotency_key, str(exc))
+            if outage is not None:
+                return outage
+        else:
+            if claim.state is ClaimState.REPLAY:
```
brandonkindred, line 370: **[LOW] standards** — The claim state handling in prepare_invoke does not have an else clause for unexpected ClaimState values. If a new state is added to the ClaimState enum, the code would silently proceed with claim_token=None, potentially bypassing the ledger. While currently exhaustive, adding an explicit else that raises an error would make the code more robust against future changes.

**Suggested fix:** Add an else clause after the REPLAY and CONFLICT checks that raises a RuntimeError for an unhandled claim state, e.g.:
```python
if claim.state is ClaimState.REPLAY:
    ...
elif claim.state is ClaimState.CONFLICT:
    ...
elif claim.state is ClaimState.CLAIMED:
    claim_token = claim.claim_token
else:
    raise RuntimeError(f"Unhandled claim state: {claim.state}")
```

backend/agents/agent_cognition/invoke_gate.py
```diff
@@ -0,0 +1,1009 @@
...
+            prepared.agent_id,
+            prepared.source_run_id,
+            prepared.claim_token,
+            status="completed",
+            response={"status_code": upstream_status, "content": content},
+        )
+        if prepared.claim_token is not None
+        else asyncio.sleep(0)
+    )
+    persisted, _ = await asyncio.gather(persist_write, complete_write)
```
brandonkindred: **[LOW] standards** — The asyncio.gather call in finalize_invoke does not use return_exceptions=True. While the inner functions (_persist_events_best_effort and _complete_best_effort) catch all exceptions, if a future change introduces an unhandled exception in one task, the other task may be abandoned without completing. Using return_exceptions=True would ensure both tasks always run to completion.

**Suggested fix:** Change the gather call to:
```python
results = await asyncio.gather(persist_write, complete_write, return_exceptions=True)
persisted = results[0] if not isinstance(results[0], BaseException) else 0
```

backend/agents/agent_cognition/invoke_gate.py
```diff
@@ -0,0 +1,1009 @@
...
+
+
+def _json_default(obj: Any) -> Any:
+    """``json.dumps`` fallback: dump nested Pydantic models, stringify the rest."""
+    if isinstance(obj, BaseModel):
+        return obj.model_dump(mode="json")
+    return str(obj)
+
+
+def _jsonable(value: Any) -> Any:
```
brandonkindred, line 1015: **[INFO] standards** — The _jsonable function uses json.loads(json.dumps(...)) to force JSON compatibility, which performs double serialization/deserialization. For large outputs this could be a performance concern, though in practice in-process outputs are small.

**Suggested fix:** Consider a more efficient approach, such as a recursive walk that converts Pydantic models to dicts and falls back to str only for non-serializable leaf values, avoiding the double conversion. This is a minor optimization and not required for correctness.

backend/agents/agent_cognition/invoke_gate.py
```diff
@@ -0,0 +1,1009 @@
...
+        "memory_events": [e.model_dump(mode="json") for e in writeback.events] if writeback else [],
+        "tool_audit": [],
+    }
+    return await finalize_invoke(prepared, 200, envelope)
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+def _storage_outage_outcome(
```
brandonkindred, line 665: **[INFO] standards** — The _storage_outage_outcome function returns None to signal 'proceed unledgered'. This is an implicit sentinel that relies on the caller checking for None. A more explicit design (e.g., returning a dedicated enum or raising a specific exception) would improve readability, but the current approach is documented and functional.

**Suggested fix:** Consider returning a dedicated sentinel value or using a separate function for the degrade path to make the control flow more explicit. For example, a PROCEED_UNLEDGERED enum member could be returned and handled in prepare_invoke.

brandonkindred (commented):
The implementation is production-quality: it correctly implements the inject-on-invoke / writeback-on-return lifecycle with idempotency, durable rule blocks, attempt-scoped memory keys, budgeted rollup catch-up, and fail-closed postconditions. The code is well-structured, follows Design by Contract and SOLID principles, and has comprehensive tests achieving 100% line coverage on the gate module. Integration with the existing route is clean, and non-cognition agents are unaffected. Minor observations are noted for future hardening but do not block approval.

**Spec compliance:** The code fully satisfies the acceptance criteria for Step 10. All specified features are implemented: source_run_id derivation, ledger claim/replay/conflict/reclaim, requires_idempotency_key enforcement, lazy rollup catch-up with budget, precondition and postcondition gates with correct persistence ordering, envelope re-cap, writeback persistence, trusted audit preservation on violation, attempt-scoped event keys, claim release, and in-process helper. The warming 202 path correctly precedes the gate. Tests cover the entire acceptance matrix.

backend/agents/agent_cognition/invoke_gate.py
brandonkindred: Fixed in 8df06a9 — the claim handling is now an exhaustive `elif` chain ending in `else: raise RuntimeError(f"Unhandled claim state: ...")`. Agreed this was worth guarding: a future `ClaimState` falling through would have proceeded with `claim_token=None`, i.e. an unledgered run silently bypassing run-once — same failure class as the unmapped-`GateOutcomeKind` guard on the response side. Regression test added (`test_prepare_unhandled_claim_state_fails_loudly`).

---

backend/agents/agent_cognition/invoke_gate.py
brandonkindred: Fixed in 8df06a9 — the gather now passes `return_exceptions=True`, logs a warning (with traceback) for either write that raises, and coerces a non-int persist result to `0`. Regression test added (`test_finalize_one_write_raising_does_not_abandon_the_other`): the persist helper is made to raise and the test asserts the sibling ledger completion still lands and the response is not masked.

---

backend/agents/agent_cognition/invoke_gate.py
brandonkindred: Keeping the round-trip, with reasoning: beyond coercion it also *validates* JSON-serializability in one step and handles the awkward corners a hand-rolled walk tends to miss (non-string dict keys → strings, tuples/sets → lists, datetimes via the default hook) with stdlib-tested semantics. A recursive walk reimplementing all of that is more code to keep correct for a path that only runs for `invoke_in_process` outputs (the HTTP path stores `upstream.json()` directly, no round-trip). If a large-output in-process consumer ever shows up in a profile, swapping the internals behind `_jsonable` is a contained change. Happy to revisit then.

---

backend/agents/agent_cognition/invoke_gate.py
brandonkindred: Keeping `GateOutcome | None`, with reasoning: the None-degrade contract is stated in the function's DbC postconditions and both call sites read as `outage = ...; if outage is not None: return outage` — the control flow is visible at the seam. Promoting the degrade path to a `PROCEED_UNLEDGERED` GateOutcomeKind would put a pseudo-outcome into the enum that is *not* a response: both transports' exhaustiveness guards (the route's and `outcome_as_finalized`'s) would then need a special case for a kind that must never be mapped, which widens exactly the surface those guards exist to keep closed. The unledgered-ness is already carried structurally where it matters — `PreparedInvoke.claim_token is None` — so a second signal would be redundant state. Happy to revisit if a third call site makes the Optional pattern harder to follow.

---

PR #847, opened 2026-06-10
Add live code-review progress, current-activity reporting, and stall detection

### LINKED ISSUE #846: No UI visibility into code review agent progress — progress bar and status freeze during reviews
## Problem

As a user, the UI shows no updates or progress while the software engineering team's code review agent is running, making it impossible to tell whether the agent is actually working, where it is in its process, or whether it has hung or silently failed.

## User stories

- As a user, I can see how far the code review agent is through its process via the job progress bar.
- As a user, I can see what the code review agent is doing at any given time.
- As a user, I can tell from the UI that the agent is running as expected and has not hung or silently failed.

## Root causes

- `CodeReviewAgent.run()` and its chunked coordinator (`backend/agents/software_engineering_team/code_review_agent/`) make one or more long LLM calls with no progress reporting — even an N-chunk review reports nothing per chunk.
- The coding team Tech Lead review (`backend/agents/coding_team/tech_lead_agent/agent.py`) silently retries up to 3 attempts with exponential backoff.
- On the default runtime path, the coding team orchestrator emits exactly one `status_text` update before each review ("Code review: {title}" / "Tech Lead reviewing: {title}") and then goes silent until the review completes.
- The `job-status` UI component does not render `status_text` at all.
- Hangs are undetectable: a 120s heartbeat thread keeps `last_heartbeat_at`/`updated_at` fresh even when the orchestrator thread is hung, so existing timestamps cannot signal a stall.

## Expected outcome

Live sub-step progress (with percentages) and a current-activity description for code reviews in the job views, plus a "last activity" indicator with a stalled warning so a hung or silently failed agent is visible from the UI.

brandonkindred:
Closes #846

## Problem

The code review agents ran as black boxes: one or more long LLM calls (with silent retries on the Tech Lead path) and zero intermediate progress reporting. The UI progress bar and status text froze for the entire review, and a hung agent was indistinguishable from a working one. More broadly, the job-level progress bar was incoherent (sub-agents wrote their own 0–100 progress straight onto the record), and nothing distinguished a hung orchestrator from a working one.

### Live progress inside the review agents
- `CodeReviewAgent.run()` and the chunked `run_coordinator` accept an optional `progress_callback(step, detail, fraction)` (`ReviewProgressCallback` in `code_review_agent/models.py`; out-of-range fractions are logged and clamped). Steps: `preparing → reviewing (chunk i/N) → parsing → finalizing → done`, non-decreasing fractions ending at 1.0. Forwarded by `quality_gate_tools.run_code_review`.
- `TechLeadAgent.run_code_review` reports each attempt and each backoff wait via a new `on_retry` hook on `llm_service.call_llm_with_retries`. Failure diagnostics report the *actual* attempt count (a fail-fast 429 previously claimed "after 3 attempts" when one ran). A raising progress callback or `on_retry` hook is logged and swallowed — it can never burn the retry budget or abort retries.

### One shared bridge into the job record
- `coding_team/activity.py` `ActivityBridge` replaces the three hand-rolled closures at the quality-gate, Tech Lead review, and `/review-pr` sites: one implementation of the `current_activity` schema (`agent/step/detail/fraction/task_id/task_title`), percent formatting (rounded, matching the UI), write coalescing (step-change or 2s minimum interval — a 20-chunk review no longer issues ~47 un-debounced PATCHes on fresh connections), and a 30s failure cooldown (a down job store cannot serialize client-retry sleeps into the review worker thread). Cleared on every exit path; the Tech Lead site's diff/evidence prep now sits inside the try so no exception can skip the clear.
- Legacy backend/frontend-code-v2 review phases bridge through `shared/review_progress.call_code_review_agent` (now also accepts `**kwargs` run signatures).

### Stall detection that works in both directions
- **`last_activity_at` is stamped centrally by the job service** on every real update (`job_service/db.py` `update_job`/`apply_patch`/`append_event`, plus a creation baseline) — every team and phase gets stall detection, not just the coding phase. `heartbeat()` deliberately never touches it; an explicit `None` is replaced, never stored. Bulk terminal markers (`mark_*_failed`/`interrupted`) clear `current_activity`.
- Both status endpoints return **`server_time`**; the UI computes activity ages against the backend clock — immune to browser clock skew in both directions.
- `staleness.util.ts` reads **only** `last_activity_at` (the old `updated_at` fallback was heartbeat-contaminated: it rendered hung jobs as freshly active and structurally suppressed the warning). Threshold raised to **10 minutes** (a single LLM call routinely runs 5–10 minutes; the client's internal 429 backoff sleeps 300s) and pending jobs are included. The SE path no longer passes a raw `get_llm` to the coding team, so the reasoning-stream flush refreshes activity during long calls.
- Stale activity can't masquerade as live progress: the terminal update, `/resume`, `/retry-failed`, the failure handlers, and `_record_failure` (which also resets the frozen percentage `status_text`) all clear `current_activity`, and the UI renders the activity card only while the job is `running`.

### A monotone job progress bar
- The SE orchestrator owns band allocation: product analysis 0–15, planning 15–30, coding 30–95 (`_scale_progress` in module-level updater factories rescales each sub-agent's own 0–100 reports). Previously PRA and Planning V3 wrote raw progress (including 100) straight onto the record, so the bar sprinted to 100 and collapsed at every handoff.
- The coding team takes its band via `run_coding_team_orchestrator(progress_base=, progress_span=)` (standalone jobs get the full 0–95), derives the value from the merged/failed task share on every snapshot persist, and no longer writes an unconditional base value — a resume can't regress the bar (the old code went 52 → 10 → 52).

### API + UI
- SE `GET /run-team/{job_id}`: `current_activity` (strictly typed but defensively coerced — a malformed stored value degrades to null instead of 500ing the endpoint), `last_activity_at`, `updated_at`, `last_heartbeat_at`, `server_time`. Coding-team `GET /status/{job_id}`: the same plus `progress` (coerced/clamped).
- One shared `StallWarningComponent` (the banner markup/SCSS had already diverged between the two inline copies) renders "Last activity: Ns ago" plus a grammatical "No agent activity for 12m — the job may be stalled." banner (previously "for 12m ago"). Polling stops on `completed_with_failures` (it ran forever). `job-status` renders `status_text`; backend/frontend-code-v2 views poll every 30s.

## Tests

Backend: 2,442 passing across the coding-team, SE, llm_service, and unified_api suites — including new coverage for `ActivityBridge` (schema, coalescing, cooldown, clear-always), central stamping contract tests (update/patch/event stamp; heartbeat never; explicit-None replaced; bulk markers clear activity), resume non-regression of the progress band, band scaling + updater factories, fail-fast attempt counts, raising-callback/hook immunity, fraction clamping, `**kwargs` bridge acceptance, and defensive `current_activity` coercion. Frontend: 1,775 passing (96%+ coverage), lint and production build green.

Pre-existing failures not touched by this change: `llm_service/tests/test_strands_provider.py` (import error on a clean tree), `test_ollama_client.py::test_extract_json_json_repair_unescaped_quotes_in_strings`, and 5 unified_api route tests (verified failing on the unmodified tree).

https://claude.ai/code/session_01BC6LsQAv2pwvuenKwYDrgy


backend/agents/coding_team/orchestrator.py
```diff
@@ -878,16 +938,40 @@ def _run_quality_gates(
...
-                code=result.get("changes_summary", ""),
-                spec_content="",
-                task_description=task.description or task.title,
-                language="python" if agent_type == "backend" else "typescript",
-                acceptance_criteria=task.acceptance_criteria or [],
-                llm_getter=self.llm_getter,
-            )
+            # Code review — bridge the agent's sub-step reports into the job record so
+            # the UI shows live review progress instead of a frozen "Code review: ..." line.
+            def _cr_progress(step: str, detail: str, fraction: float) -> None:
```
brandonkindred: **[MEDIUM] logic** — The `_cr_progress` function is defined inside a loop over tasks and references the loop variable `task` directly. In Python, closures capture variables by reference, so if the callback is ever invoked after the loop iteration completes (e.g., if the review becomes asynchronous), it would use the wrong task. While currently the callback is called synchronously within the same iteration, this pattern is fragile and a common source of bugs.

**Suggested fix:** Use a default argument to capture the current value, similar to `_tl_progress` on line 1025. Change the definition to:
```
def _cr_progress(step: str, detail: str, fraction: float, task: Task = task) -> None:
    pct = int(fraction * 100)
    _safe_activity_update(
        update_fn,
        status_text=f"Code review ({pct}%): {task.title}"
        + (f" — {detail}" if detail else ""),
        current_activity={
            "agent": "code_review",
            "step": step,
            "detail": detail,
            "fraction": fraction,
            "task_id": task.id,
            "task_title": task.title,
        },
    )
```
(Ensure `Task` is imported or use `Any`.)

backend/agents/coding_team/api/main.py
```diff
@@ -918,12 +940,41 @@ def _run_pr_review(job_id: str, request: ReviewPrRequest, token: str) -> None:
...
+
+            def _pr_progress(step: str, detail: str, fraction: float) -> None:
+                """Surface review sub-steps on the job record (best-effort).
+
+                Preconditions: invoked by the review agent per its progress contract.
+                Postconditions: a failed store write is logged and swallowed —
+                    observability must never fail the review itself.
+                """
+                try:
+                    update_job(
```
brandonkindred: **[MEDIUM] logic** — The progress callback in the `/review-pr` endpoint updates `status_text` and `current_activity` but does not set `last_activity_at`. The new stall detection feature relies on `last_activity_at` to differentiate real activity from heartbeats. Without it, the UI will fall back to `updated_at`, which may be less accurate or also updated by heartbeats, potentially masking stalls for PR review jobs.

**Suggested fix:** Add `last_activity_at` to the `update_job` call inside `_pr_progress`. Import `datetime` and `timezone` if not already, and include `last_activity_at=datetime.now(timezone.utc).isoformat()` in the `update_job` call. Also ensure the `finally` block does not clear `last_activity_at` (it only clears `current_activity`, so it's fine).

backend/agents/coding_team/api/main.py
```diff
@@ -918,12 +940,41 @@ def _run_pr_review(job_id: str, request: ReviewPrRequest, token: str) -> None:
...
             )
+
+            def _pr_progress(step: str, detail: str, fraction: float) -> None:
+                """Surface review sub-steps on the job record (best-effort).
+
+                Preconditions: invoked by the review agent per its progress contract.
+                Postconditions: a failed store write is logged and swallowed —
+                    observability must never fail the review itself.
+                """
+                try:
```
brandonkindred: **[LOW] standards** — The PR review endpoint uses a local try/except to swallow store errors, while the orchestrator uses a shared `_safe_activity_update` helper. This duplicates the error-handling logic and could lead to inconsistencies if the safe-update behavior needs to change.

**Suggested fix:** Extract `_safe_activity_update` to a shared utility module (e.g., `coding_team.shared.progress_utils`) and reuse it in the PR review endpoint to keep the "observability never aborts" contract in one place.

backend/agents/coding_team/orchestrator.py
```diff
@@ -445,9 +489,23 @@ def run_coding_team_orchestrator(
...
+
+        The 120s heartbeat thread keeps last_heartbeat_at/updated_at fresh even when the
+        orchestrator thread is hung, so those timestamps cannot signal a stall; this field
+        advances only on genuine orchestrator activity and is what the UI's stall warning reads.
+
+        Preconditions: ``kw`` values are JSON-serializable.
+        Postconditions: the job record carries a UTC ISO ``last_activity_at``; an explicit
+            ``last_activity_at`` in ``kw`` is preserved (setdefault semantics).
+        """
+        kw.setdefault("last_activity_at", datetime.now(timezone.utc).isoformat())
```
brandonkindred: **[LOW] logic** — The line `kw.setdefault("last_activity_at", datetime.now(timezone.utc).isoformat())` will not overwrite an explicit `last_activity_at=None` passed by a caller. If a caller ever passes `None` (e.g., to clear the field), it would remain `None`, potentially breaking the UI's stall detection. Currently no caller does this, but it's a latent bug.

**Suggested fix:** Change to `if "last_activity_at" not in kw or kw["last_activity_at"] is None: kw["last_activity_at"] = datetime.now(timezone.utc).isoformat()` to ensure a valid timestamp is always set unless explicitly overridden with a string.

backend/agents/software_engineering_team/code_review_agent/models.py
```diff
@@ -1,11 +1,39 @@
...
+
+Steps: ``preparing | reviewing | waiting_retry | parsing | finalizing | done``.
+
+Preconditions (on the callable a caller provides):
+    - Must not raise; review progress is observability, never control flow.
+    - Must accept (str, str, float) positionally.
+"""
+
+
+def notify_review_progress(
```
brandonkindred, line 24: **[INFO] standards** — The `notify_review_progress` function does not validate that `fraction` is between 0.0 and 1.0. While the callers are expected to provide valid values, an accidental out-of-range fraction could cause UI issues (though the UI clamps). Adding a debug assertion would help catch bugs during development.

**Suggested fix:** Add `assert 0.0 <= fraction <= 1.0, f"Invalid fraction: {fraction}"` at the start of the function, or log a warning if outside range.

brandonkindred (commented):
The PR delivers a well-structured implementation of live progress reporting, current-activity tracking, and stall detection. The code follows Design by Contract with clear pre/post conditions, maintains backward compatibility, and includes extensive tests. The main areas for improvement are: (1) a potential closure late-binding issue in the orchestrator's `_cr_progress`, (2) missing test coverage for the PR review progress callback, and (3) the PR review flow not setting `last_activity_at`, which slightly weakens stall detection for that path. These are medium severity and should be addressed but do not block merge.

**Spec compliance:** The implementation fully meets the acceptance criteria: progress callbacks are integrated into the review agents, bridged to job records, and surfaced in the UI with sub-progress bars and stall warnings. The overall job progress bar now advances during the coding phase. All specified steps (preparing, reviewing, parsing, finalizing, done) are reported with non-decreasing fractions. The stall detection correctly uses `last_activity_at` to avoid heartbeat interference. The legacy review phases are bridged via the shared `call_code_review_agent` helper with signature guarding. The only gap is the missing `last_activity_at` in the PR review flow, which is a minor deviation from the full stall-detection design.

- `backend/agents/coding_team/tests/test_review_pr.py` — **[MEDIUM] testing** — The PR review endpoint (`/review-pr`) now includes a progress callback that updates `status_text` and `current_activity` on the job. However, the existing tests in `test_review_pr.py` do not verify that these updates are actually written to the job store. This leaves the new progress reporting untested.

**Suggested fix:** Add a test that mocks `update_job` and asserts that it is called with the expected `status_text` and `current_activity` during the review, and that `current_activity` is cleared afterward. For example, use `unittest.mock.patch` to capture calls to `update_job` and verify the arguments.

backend/agents/coding_team/orchestrator.py
brandonkindred: Fixed in d683eca, by removing the closure entirely rather than adding a default-arg capture: the three hand-rolled progress closures (`_cr_progress`, `_tl_progress`, `_pr_progress`) are replaced by one shared `ActivityBridge` (`coding_team/activity.py`) constructed per task inside the loop, with `task_id`/`task_title` bound at construction time — no loop-variable capture remains. The bridge also centralizes the percent formatting, write coalescing, and swallow-on-failure behavior.

---

backend/agents/coding_team/api/main.py
brandonkindred: Fixed in d683eca at a deeper level than a per-callsite stamp: `last_activity_at` is now stamped centrally by the job service on every real update (`job_service/db.py` — `update_job`, `apply_patch`, `append_event`, plus a creation-time baseline), while `heartbeat()` deliberately never touches it. So the `/review-pr` flow's writes (and every other team/phase, including SE planning, which previously never stamped at all) count as activity automatically. The UI also no longer falls back to `updated_at` — that fallback was heartbeat-contaminated and would have masked stalls exactly as you describe.

---

backend/agents/coding_team/api/main.py
brandonkindred: Done in d683eca: the duplicated swallow logic is consolidated into `coding_team/activity.py`'s `ActivityBridge`, used by the orchestrator's quality-gate and Tech Lead review sites and the `/review-pr` endpoint. It owns the "observability never aborts" contract in one place, and additionally coalesces same-step reports (one write per 2s or on step change) and opens a 30s cooldown after a failed write, so a down job store can't serialize client-retry sleeps into the review thread once per report. The old `_safe_activity_update` helper is removed.

---

backend/agents/coding_team/orchestrator.py
brandonkindred: Fixed in d683eca. The client-side `setdefault` is gone entirely — stamping moved into the job service (`job_service/db.py:update_job`) so there's one clock and one writer — and the central implementation uses exactly the semantics you suggested: `if fields.get("last_activity_at") is None: fields["last_activity_at"] = now`, so an explicit `None` is replaced with a fresh stamp and the field can never go invalid once the job exists. Covered by `test_fake_update_job_stamps_activity` in the job-service client contract tests.

---

backend/agents/software_engineering_team/code_review_agent/models.py
brandonkindred: Done in d683eca: `notify_review_progress` now validates the range and logs a warning + clamps to [0.0, 1.0] rather than asserting — an out-of-range fraction is a reporter bug, but progress reporting is observability and must not abort a review over a bad number. Test: `test_notify_review_progress_clamps_out_of_range_fraction`.

---

brandonkindred:
@codex review

backend/agents/coding_team/orchestrator.py
```diff
@@ -472,6 +511,7 @@ def _persist_graph() -> None:
             agent_task_map=snap["agent_task_map"],
             phase=phase,
             status_text=status_text,
+            progress=_coding_progress(snap["tasks"], progress_base, progress_span),
```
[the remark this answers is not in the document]
brandonkindred: Fixed in 1e8be17. `execute_coding_team_activity` now passes `PROGRESS_BAND_CODING` (30–65) through `progress_base`/`progress_span`, and the Temporal PRA/planning updaters were replaced with the same `_make_pra_job_updater`/`_make_planning_v3_job_updater` band-scaling factories the thread path uses — passing the coding band alone wouldn't have fixed the collapse, since planning was still writing its raw 100 onto the record. Both halves are pinned by `test_execute_coding_team_activity_passes_band_and_default_llm_getter` and `test_temporal_pra_and_planning_updaters_are_the_shared_band_factories`.

While fixing this, it turned out the Temporal planning→coding handoff had never actually worked: `PlanningV2AdapterResult` is a dataclass, so `plan_project_activity`'s `hasattr(model_dump)` probe silently serialized `{}` and `execute_coding_team_activity`'s `model_validate` call raised unconditionally. The dataclass now has explicit `to_dict`/`from_dict` (nested Pydantic models handled), used by both activities and covered by a lossless round-trip test.

---

backend/agents/software_engineering_team/orchestrator.py
```diff
@@ -2950,13 +3018,20 @@ def _run_architecture_for_planning_v3(  # pragma: no cover  # integration-only:
             )
             from coding_team.orchestrator import run_coding_team_orchestrator
 
+            base, span = PROGRESS_BAND_CODING
+            # get_llm deliberately NOT passed: the coding team's default getter wraps
+            # the LLM clients with reasoning-stream capture, whose periodic flush is
+            # the only thing that refreshes job activity DURING a multi-minute LLM
+            # call — passing the raw get_client here made every long implement call
+            # look like a stall to the UI's activity-based warning.
```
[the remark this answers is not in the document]
brandonkindred: Fixed in 1e8be17 — `get_llm=get_client` is dropped from the Temporal call, so the coding team's default reasoning-stream getter (with its periodic thinking flushes refreshing `last_activity_at`) now applies on both paths.

One correction to the severity, for the record: this was worse than a false stall warning. `get_client` returns a raw `OllamaLLMClient`, which doesn't implement the strands `Model` ABC, and `TechLeadAgent.__init__` does `Agent(model=...)` — under `LLM_PROVIDER=ollama` that construction raises `AttributeError: 'OllamaLLMClient' object has no attribute 'stateful'` (reproduced against strands 1.42 in the repo venv), so the Temporal coding phase failed outright at startup; only `LLM_PROVIDER=dummy` survived, which is why CI never saw it. Dropping the raw getter fixes both the crash and the activity-refresh gap. Pinned by `test_execute_coding_team_activity_passes_band_and_default_llm_getter`.

---

brandonkindred:
@codex review

brandonkindred:
@codex review

backend/agents/software_engineering_team/code_review_agent/models.py
```diff
@@ -1,11 +1,50 @@
...
+          exactly once.
+    """
+    if callback is None:
+        return
+    if not 0.0 <= fraction <= 1.0:
+        logging.getLogger(__name__).warning(
+            "review progress fraction out of range (step=%s): %r; clamping", step, fraction
+        )
+        fraction = min(max(fraction, 0.0), 1.0)
+    callback(step, detail, fraction)
```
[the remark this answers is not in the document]
brandonkindred: Fixed in cb42db6. `notify_review_progress` now catches and logs callback exceptions at the invocation boundary, so a raising callback can never abort a review or divert it to the LLM fallback — consistent with the existing Tech Lead `_report` and `call_llm_with_retries` `on_retry` guards. Since every agent- and coordinator-side report flows through `notify_review_progress`, this one guard covers all of `CodeReviewAgent.run()` and the chunked coordinator.

For the record on reachability: every production callback today already swallows internally (the coding-team sites go through `ActivityBridge`, whose `__call__` never raises, and the v2 bridge's `detail_callback` chain ends in the team orchestrators' `_update_job`, which try/excepts everything) — so this was a latent hazard for future/external callbacks rather than reachable today. Still worth closing: a misbehaving reporter now costs a logged warning instead of silently replacing the configured reviewer's findings. The callback contract docstring is updated, and `test_raising_callback_is_swallowed_and_never_changes_the_review` pins both halves (result unchanged, warning logged).

---

brandonkindred (commented):
The PR implements live progress reporting, current-activity tracking, and stall detection across the code review agents, orchestrator, and UI. The code follows SOLID principles, uses Design by Contract, and includes comprehensive tests. No critical or high issues found.

**Spec compliance:** The implementation meets the acceptance criteria: progress callbacks are forwarded, activity bridge coalesces writes, stall detection uses last_activity_at centrally stamped, UI shows sub-agent progress and stall warnings. All specified features are present.

_Note: this review covered the first 50 changed file(s); 2 further changed file(s) exceeded the per-review cap (`PR_REVIEW_MAX_FILES`) and were not inspected._

brandonkindred:
@codex review

PR #867, opened 2026-06-12
Make the engine own exits for custom-code Strategy Lab strategies

### LINKED ISSUE #866: Strategy Lab: custom-code strategies hand-roll exits the engine already owns, stripping engine_exit attribution
## Problem

Custom-code (LLM-synthesized) Strategy Lab strategies submit their own position-closing orders for exit conditions. But the execution engine **already** enforces `spec.exit_rules` for the custom-code path too — the backtest mode passes `exit_rules=list(strategy.exit_rules)` unconditionally, and the per-run dispatcher evaluates stop-loss / take-profit / signal-exit rules every bar.

Because the strategy's manual close is queued *before* the engine's close, it fills first and the resulting trade record's exit reason is stamped from the strategy order (empty) instead of `engine_exit:<kind>`. This:

1. **Strips the engine attribution** the deterministic trade-alignment gate relies on, and
2. **Drifts from the spec's structured exit semantics** (bar timing, intrabar-high vs close, cross definition), so the gate's recomputation of the spec predicate at the exit bar fails.

Observed quality-gate failures:

```
[signal_exit] Trade #24 exited without engine attribution, but no
SignalExitRule predicate fires at the exit bar.

trade_alignment: manual exit orders for the MACD cross-under are never
attributed to an engine-side signal exit rule, and the take-profit cap
is not enforced because the engine's take-profit mechanism is not engaged.
```

The deterministic compiler already avoids this with "engine-only ordering" (it submits no orders); the custom-code path was doing the opposite for exits, recreating the dual-path conflict the compiler was built to eliminate. The root cause was reinforced by the conformance gate, which *required* custom code to author a manual signal-exit branch.

## Fix

Align the custom-code path with the engine-only-exit design (entries stay inline, since the engine entry dispatcher is off for custom code):

- **Prompts** — instruct the synthesis and alignment-repair models that the engine owns every `spec.exit_rules` entry; author entries only, never submit a closing order for a spec-declared exit, and remove such closes during repair.
- **Code-conformance gate** — retire the requirement that custom code implement a manual signal-exit branch; add a coverage-gated hard critical forbidding any position-closing `submit_order` when the engine's exit rules cover the entered sides; stop requiring the code to read indicators referenced only by engine-owned exit rules.
- **Predicate-conformance gate** — skip signal-exit fixtures when the engine covers the entered sides (they would otherwise demand a manual close the new contract forbids).

brandonkindred:
Closes #866

## Why

Custom-code (LLM-synthesized) strategies were hand-rolling exit orders that duplicate the structured exit rules the engine **already** enforces on every run (the backtest mode passes `exit_rules` to the engine unconditionally). The manual close fills ahead of the engine's close and overwrites the `engine_exit:<kind>` attribution the trade-alignment gate depends on — producing spurious `signal_exit` / `take_profit` misalignment criticals (a MACD cross-under exit attributed to no engine rule; a take-profit cap that reads as unenforced).

The deterministic compiler already avoids this with "engine-only ordering". This PR extends that contract to the custom-code path **for exits** — entries stay inline, since the engine entry dispatcher is off for custom code.

## What

**Prompts**
- `code_synthesis_system.md` — remove the manual `# ── EXIT ──` branch and the "closing a position" guidance; the engine owns every `spec.exit_rules` entry and stamps `engine_exit:<kind>`; the strategy authors entries only.
- `alignment_propose_fix.md` — the repair path now **removes** a manual close that duplicates a spec-declared exit rather than adding/strengthening one.

**`code_conformance.py`**
- Retire the requirement that custom code implement a manual signal-exit branch (`SignalExitRule`s are engine-owned for both paths).
- New coverage-gated hard critical (`_check_no_duplicate_engine_exit`): forbids any position-closing `submit_order` when `_engine_exits_cover_sides` holds for the entered sides. Skips the compiled path; allows manual closes for sides the engine can't cover (preserves legitimate bespoke exits).
- Stop requiring the code to read indicators referenced **only** by engine-owned exit rules.

**`predicate_conformance.py`**
- Skip signal-exit fixtures when the engine covers the entered sides; otherwise the shadow check would demand a manual close the new contract forbids. Entry fixtures still run.

## Design decisions
- **Coverage-gated** prohibition (reuses the existing `_engine_exits_cover_sides` helper) so partial-coverage specs don't false-positive.
- Shipped as a **hard critical** (consistent with the other conformance criticals).
- The duplicate-exit matcher only flags the literal `qty=position.qty` close shape — `**kwargs`-spread closes are intentionally not flagged (conservative).

## Testing
- `test_code_conformance_gate.py`: inverted the old "requires manual exit branch" test; added the new duplicate-exit critical plus its compiled-path, partial-coverage, and kwargs-spread carve-outs.
- `test_predicate_conformance_gate.py`: added a positive coverage-skip test (would false-negative without this change).
- Full `investment_team` suite: **3407 passed, 10 skipped**. `ruff check` + `ruff format --check` clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01HUzHra8eV7rDL7PqsAXw8B)_


backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -807,61 +819,128 @@ def _check_entry_coverage(self, cctx: _ConformanceCtx) -> Iterable[QualityGateRe
...
+        exit_rules = getattr(spec, "exit_rules", None) or []
+        if not exit_rules:
+            return ()
+        # Entered sides come from the spec (the authoritative, validated
+        # source) rather than the AST: a dynamic ``side=`` expression would
+        # leave an AST-derived side set empty and silently disable the check.
+        sides = {
+            r.side for r in (getattr(spec, "entry_rules", []) or []) if isinstance(r, EntryRule)
+        }
+        if not sides or not _engine_exits_cover_sides(spec, sides):
```
brandonkindred: **[HIGH] logic** — The duplicate-exit check only forbids manual closes when _engine_exits_cover_sides returns True for ALL entered sides. If some sides are covered but not all, it allows manual closes for all sides, including those that are engine-owned. This contradicts the engine-owned exits contract and can cause exit misalignment.

**Suggested fix:** Change the condition to forbid manual closes for any side that is covered. For each submit_order close, determine the side it closes and check if that side is covered by any engine exit rule. Alternatively, if the design intends that partial coverage means strategy owns all exits, the engine must be configured to not enforce its rules for custom-code strategies when coverage is partial; but the PR states the engine enforces unconditionally. So the check should be per-side.

backend/agents/investment_team/strategy_lab/quality_gates/predicate_conformance.py
```diff
@@ -313,6 +325,16 @@ def check(
...
             fixtures = generate_conformance_fixtures(spec, compiled_code=code)
+            # Signal exits are engine-owned: when spec.exit_rules cover every
+            # entered side, the engine enforces the SignalExitRule via
+            # _EngineExitDispatcher and conforming custom code submits no manual
+            # close. The shadow context only sees strategy-submitted orders, so
+            # keeping those fixtures would demand a close the conformance
+            # contract now forbids. Drop them; entry fixtures still run
+            # (entries remain inline for custom code).
+            entered_sides = {r.side for r in entry_rules if isinstance(r, _EntryRule)}
+            if entered_sides and _engine_exits_cover_sides(spec, entered_sides):
```
brandonkindred, line 338: **[HIGH] logic** — The predicate conformance gate skips signal-exit fixtures only when ALL entered sides are covered. If coverage is partial, it keeps all signal-exit fixtures, demanding manual closes for signal exits on sides that are engine-owned. This conflicts with the engine-owned exits contract and can cause the synthesis loop to demand a manual close that the code conformance gate would forbid (or allow but cause misalignment).

**Suggested fix:** Filter out signal-exit fixtures individually based on whether the specific rule's side is covered by the engine, rather than using an all-sides condition. For each fixture with rule_kind == 'signal_exit', determine its side and check if that side is covered; skip if covered.

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -106,14 +107,26 @@ def _indicators_in_predicate(p: Predicate) -> set[str]:
...
+    compute. Entry-rule indicators are always included — entries are
+    authored inline on both the compiled and custom-code paths.
+    ``SignalExitRule`` indicators are included only for the compiled path:
+    on the custom-code path exits are engine-owned (the engine computes
+    their indicators via ``_EngineExitDispatcher`` and the strategy
+    authors no exit branch), so requiring the code to read an exit-only
+    indicator would contradict the engine-owned-exits contract.
+    """
     refs: set[str] = set()
     for rule in getattr(spec, "entry_rules", []) or []:
```
brandonkindred, line 131: **[MEDIUM] logic** — The function _collect_required_indicators uses getattr(spec, ...) without checking if spec is None. The docstring says spec can be None, but calling getattr(None, ...) raises AttributeError. Other methods like _check_signal_exit_coverage and _check_no_duplicate_engine_exit have the same issue.

**Suggested fix:** Add a guard at the start of each method: `if spec is None: return ()` (or appropriate default). Alternatively, update docstrings to reflect that spec is always a StrategySpec and ensure callers never pass None.

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -807,61 +819,128 @@ def _check_entry_coverage(self, cctx: _ConformanceCtx) -> Iterable[QualityGateRe
...
-                self._critical(
-                    f"Spec declares {len(signal_exits)} signal-exit rule(s) "
-                    f"but only {len(exit_branches)} exit branch(es) were found "
-                    "reachable from on_bar."
-                ),
-            )
-        return ()
+        seen: set[int] = set()
+        n_closes = 0
+        for _if_node, calls in cctx.submit_branches:
```
brandonkindred: **[MEDIUM] logic** — The duplicate-exit check only scans submit_order calls inside if branches (via cctx.submit_branches). A submit_order call that closes a position placed outside any conditional block (e.g., at the top level of on_bar) would be missed, allowing a duplicate engine exit to pass undetected.

**Suggested fix:** Extend the check to also scan top-level statements in the on_bar method for submit_order calls that close positions. This could be done by adding a helper that collects all submit_order calls in the method body, or by ensuring _find_if_branches_with_submit_order captures top-level calls as well.

brandonkindred (commented):
The PR successfully updates prompts and adds new conformance checks to enforce engine-owned exits for custom-code strategies. However, the logic for partial coverage (where some but not all entered sides are covered by engine exit rules) is inconsistent between the code conformance and predicate conformance gates, and the duplicate-exit check is too permissive, allowing manual closes for engine-owned sides when coverage is partial. This can lead to exit misalignment and synthesis loop failures. Additionally, there are medium issues with None-safety, incomplete AST scanning, and missing test coverage. These should be addressed before merge.

**Spec compliance:** The code meets the acceptance criteria for the task (engine owns exits, prompts updated, new checks added). However, the partial coverage edge case is not correctly handled, which may cause failures in strategies with mixed coverage. The spec did not explicitly require handling partial coverage, but the design decision mentions coverage-gated prohibition to avoid false positives; the current implementation's all-or-nothing approach is a flaw.

- `backend/agents/investment_team/tests/test_code_conformance_gate.py` — **[MEDIUM] testing** — Missing test for partial coverage scenario where some entered sides are covered by engine exits but others are not. The current tests only cover all-covered or none-covered cases. This leaves the inconsistent behavior untested and could allow regressions.

**Suggested fix:** Add a test with two entry sides (e.g., long and short), exit rules covering only long, and a manual close for long. Expect the duplicate-exit check to fire (if per-side logic is implemented) or at least document expected behavior.
- `backend/agents/investment_team/tests/test_predicate_conformance_gate.py` — **[MEDIUM] testing** — Missing test for partial coverage where a signal-exit rule exists for a covered side but another side is uncovered. The current test only covers the all-covered case.

**Suggested fix:** Add a test with multiple entry sides, partial coverage, and a signal-exit rule for a covered side; verify that the fixture is skipped (if per-side filtering is implemented) or that the behavior is consistent with the code conformance gate.

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -807,61 +819,128 @@ def _check_entry_coverage(self, cctx: _ConformanceCtx) -> Iterable[QualityGateRe
...
+            for call in calls:
+                if id(call) in seen:
+                    continue
+                if _submit_order_closes_position(call):
```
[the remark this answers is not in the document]
brandonkindred: Fixed in `34fd5be`. The check now scans **every** `on_bar`-reachable `submit_order` via `_iter_method_body_nodes` (if/elif/else bodies, post-guard top-level statements, and reachable helpers) instead of the if-branch-only `submit_branches` view, so a close in a plain `else` (the old template shape) is caught. Added a regression test `test_no_duplicate_engine_exit_detects_close_in_else_block`.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
brandonkindred: Fixed in `34fd5be`. Coverage is now judged **per-side**: each close's `side=` literal is mapped back to the position side it retires (`_submit_order_close_side` — a `SHORT`/`sell` order closes a long, a `LONG`/`buy` order closes a short), and the close is forbidden only when that specific side is engine-owned. A close whose side can't be resolved statically is checked against every entered side (forbid if any is engine-owned). Added `test_no_duplicate_engine_exit_partial_coverage_flags_covered_side` and `..._allows_uncovered_side`.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
brandonkindred: Fixed in `34fd5be`, together with the `else`-block case — the check now walks every `on_bar`-reachable `submit_order` (top-level statements included) via `_iter_method_body_nodes`, not just calls inside `if` branches.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
brandonkindred: Added explicit `if spec is None` guards to `_collect_required_indicators`, `_check_signal_exit_coverage`, and `_check_no_duplicate_engine_exit` in `34fd5be`. One clarification: `getattr(None, "x", default)` returns the default rather than raising `AttributeError` (only the two-arg form raises), and `_engine_exits_cover_sides` already short-circuits on `None`, so the prior code was already None-safe in practice — but the explicit guards make the contract self-evident.

---

backend/agents/investment_team/strategy_lab/quality_gates/predicate_conformance.py
brandonkindred: A `SignalExitRule` is engine-owned for **both** sides, so partial coverage can't arise for a signal-exit fixture: `_engine_exits_cover_sides`'s per-rule check returns `True` for a `SignalExitRule` regardless of side (`code_safety.py:468-469`), and the engine's `evaluate_exit_rules` fires it for any open position irrespective of side. Whenever a signal-exit fixture exists (the spec carries a `SignalExitRule`), every entered side is therefore covered and `_engine_exits_cover_sides(entered_sides)` is `True` — so the all-sides gate is already equivalent to per-side here. The genuine partial-coverage case (handled per-side in the code-conformance gate as of `34fd5be`) only arises for side-specific stops (`trailing_high`/`trailing_low`), which aren't predicate fixtures. Added `test_signal_exit_skipped_even_with_side_specific_stop` to lock this in. I left this thread open in case you'd prefer the predicate filter rewritten in an explicit per-fixture form, but given the both-sides coverage it would be a behavioral no-op.

---

backend/agents/investment_team/strategy_lab/prompts/code_synthesis_system.md
```diff
@@ -72,9 +72,12 @@ class MyStrategy(Strategy):
             pass
 
         # ── EXIT ──────────────────────────────────────────
```
brandonkindred: **[MEDIUM] spec-compliance** — The prompt still contains the section header "# ── EXIT ──" even though the code branch has been removed. This header may mislead the LLM into thinking it should place exit logic in that section, despite the following comment instructing otherwise. The specification required removing the manual "# ── EXIT ──" branch; while the code is removed, the header remains and could cause confusion.

**Suggested fix:** Remove the "# ── EXIT ──" line entirely, or replace it with a non-actionable header such as "# ── ENGINE OWNS EXITS (DO NOT IMPLEMENT) ──" to eliminate any ambiguity.

brandonkindred (commented):
The PR successfully implements the engine-owned exits contract for custom-code strategies. The quality gates are updated to forbid manual closes that duplicate engine-owned exits, retire the requirement for manual signal-exit branches, and skip signal-exit fixtures in predicate conformance when the engine covers the entered sides. The prompts are updated to instruct the LLM accordingly. Tests cover the new behavior and edge cases. The code is well-structured, documented, and follows existing patterns. One minor issue: the code synthesis prompt retains the "# ── EXIT ──" section header which could cause confusion; it should be removed or reworded.

**Spec compliance:** The implementation meets the acceptance criteria: prompts updated, code conformance gate retired manual signal-exit requirement and added duplicate-exit prohibition, predicate conformance gate skips signal-exit fixtures when engine covers sides, indicator requirements adjusted, and tests added. The only deviation is the leftover EXIT header in the prompt, which is a minor spec-compliance nit.

brandonkindred:

backend/agents/investment_team/strategy_lab/prompts/code_synthesis_system.md
brandonkindred: Fixed in `6440f95`. The template header is now `# ── ENGINE OWNS EXITS (do NOT submit close orders) ──` followed by the explanatory comment, so there's no actionable `# ── EXIT ──` slot left to mislead.

---

backend/agents/investment_team/strategy_lab/prompts/code_synthesis_system.md
```diff
@@ -72,9 +72,12 @@ class MyStrategy(Strategy):
...
-        else:
-            # <FILL IN exit decision per spec.exit_rules>
-            pass
+        # Exits are engine-owned: the engine enforces every
+        # spec.exit_rules entry (stop-loss, take-profit, signal-exit) and
+        # stamps engine_exit:<kind> on the close. Do NOT submit a closing
+        # order for a spec-declared exit — a manual close fills first,
+        # strips that attribution, and fails the trade-alignment gate.
+        # on_bar submits entries only.
```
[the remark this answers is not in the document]
brandonkindred: Good catch — fixed in `6440f95`. The prompt no longer says "entries only" unconditionally. It now instructs: do not close a side the engine covers, but DO submit a manual close for a position side that no `spec.exit_rules` entry covers (e.g. a short when the only stop is a long-side `trailing_high`). That matches the per-side duplicate-exit check and `CodeSafetyChecker`'s order-flow relaxation, both of which require/allow a manual close for engine-uncovered sides.

---

backend/agents/investment_team/strategy_lab/prompts/alignment_propose_fix.md
```diff
@@ -61,7 +61,15 @@ A JSON object with these fields:
...
+  ``hold_count``, ``if counter >= N: close``); they are rejected by the
+  conformance gate.
+- Exits are engine-owned. The engine enforces every ``spec.exit_rules``
+  entry (stop-loss / take-profit / signal-exit) and stamps
+  ``engine_exit:<kind>`` attribution. When a finding cites a
+  ``signal_exit`` or ``take_profit`` divergence, the fix is to REMOVE
+  the strategy's manual position-closing order (opposite ``side``,
+  ``qty == position.qty``) for that exit and let the engine own it — not
+  to add or strengthen a manual close. The repaired ``on_bar`` submits
+  entries only.
```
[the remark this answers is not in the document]
brandonkindred: Fixed in `6440f95`. The repair constraint now scopes removal to manual closes that duplicate an **engine-covered** exit, and explicitly says to KEEP a manual close for a position side no exit rule covers — removing it would strand the position and fail the safety gate.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -807,61 +858,141 @@ def _check_entry_coverage(self, cctx: _ConformanceCtx) -> Iterable[QualityGateRe
...
+            r.side for r in (getattr(spec, "entry_rules", []) or []) if isinstance(r, EntryRule)
+        }
+
+        n_closes = 0
+        for method in _iter_strategy_methods(cctx.cls):
+            if method.name not in cctx.reachable:
+                continue
+            for node in _iter_method_body_nodes(method):
+                if not _is_submit_order_call(node) or not _submit_order_closes_position(node):
+                    continue
```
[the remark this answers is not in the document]
brandonkindred: Fixed in `6440f95`. The duplicate-exit check now intersects the inferred closed side with the spec's entered sides: a real close retires a side the spec actually holds, so a same-side full-size scale-in (whose inferred "closed" side — the opposite of the order side — is not an entered side) is skipped rather than flagged. That resolves the one-sided-spec false positive you called out. Added `test_no_duplicate_engine_exit_ignores_same_side_scale_in`. (The residual ambiguity — a full-size same-side scale-in in a genuinely two-sided spec — can't be resolved statically without the runtime position side; the intersect handles the decidable cases.)

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -374,6 +389,43 @@ def _submit_order_closes_position(call: ast.Call) -> bool:
...
+
+    Pre: ``call`` is a ``ctx.submit_order(...)`` ``ast.Call``.
+    Post: returns ``"long"`` / ``"short"`` when the ``side=`` value is a
+    string literal (``"SHORT"`` / ``"sell"`` / ...) or an
+    ``OrderSide.<NAME>`` attribute that maps to a known order side;
+    ``None`` when ``side`` is absent, a non-literal expression that cannot
+    be resolved statically, or an unrecognised token. The mapping inverts
+    the order side because a close submits the side opposite the position.
+    """
+    for kw in call.keywords:
```
brandonkindred, line 421: **[LOW] logic** — The function `_submit_order_close_side` only inspects `call.keywords` to locate the `side` argument. If `ctx.submit_order` is invoked with positional arguments (e.g., `ctx.submit_order(symbol, qty, side)`), the `side` value will reside in `call.args` and be missed, potentially allowing a duplicate engine exit to go undetected.

**Suggested fix:** Extend the function to also check positional arguments by examining `call.args` if the parameter order of `submit_order` is known and stable, or add a comment documenting that only keyword arguments are supported.

backend/agents/investment_team/strategy_lab/quality_gates/predicate_conformance.py
```diff
@@ -313,6 +325,16 @@ def check(
...
 
             fixtures = generate_conformance_fixtures(spec, compiled_code=code)
+            # Signal exits are engine-owned: when spec.exit_rules cover every
+            # entered side, the engine enforces the SignalExitRule via
+            # _EngineExitDispatcher and conforming custom code submits no manual
+            # close. The shadow context only sees strategy-submitted orders, so
+            # keeping those fixtures would demand a close the conformance
+            # contract now forbids. Drop them; entry fixtures still run
+            # (entries remain inline for custom code).
+            entered_sides = {r.side for r in entry_rules if isinstance(r, _EntryRule)}
```
brandonkindred: **[LOW] logic** — The set `entered_sides` is constructed from `r.side` for each `EntryRule`. If an entry rule has `side=None` (e.g., a rule that applies to both sides), the set will contain `None`. This value is then passed to `_engine_exits_cover_sides(spec, entered_sides)`, which may not handle `None` correctly, leading to incorrect skipping of signal-exit fixtures.

**Suggested fix:** Filter out `None` values from `entered_sides` before passing to `_engine_exits_cover_sides`, e.g., `entered_sides = {r.side for r in entry_rules if isinstance(r, _EntryRule) and r.side is not None}`. Alternatively, ensure `_engine_exits_cover_sides` treats `None` as a side that is never covered.

brandonkindred (commented):
The PR correctly implements the engine-owned exits contract for custom-code strategies. Prompts are updated, conformance gates are modified appropriately, and tests cover the new behavior comprehensively. A few minor edge cases in AST analysis and documentation could be improved for robustness.

**Spec compliance:** All acceptance criteria are met: prompts remove manual exit guidance and instruct removal of duplicate closes; code_conformance retires the old signal-exit requirement and adds a hard critical for duplicate engine exits; predicate_conformance skips signal-exit fixtures when the engine covers all entered sides; tests validate each scenario. The implementation aligns with the design decisions (coverage-gated, hard critical, conservative kwargs-spread handling).

- `backend/agents/investment_team/tests/test_code_conformance_gate.py` — **[LOW] testing** — There is no test case for a close with a dynamically computed side expression (e.g., `side = "SELL" if condition else "BUY"`) where the engine covers one side but not the other. This would verify the conservative behavior of the duplicate-exit check when the side cannot be statically resolved.

**Suggested fix:** Add a test that uses a non-literal side expression and confirms the check flags it when any entered side is engine-owned, and another test where all entered sides are uncovered to confirm it does not flag.

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -807,61 +858,154 @@ def _check_entry_coverage(self, cctx: _ConformanceCtx) -> Iterable[QualityGateRe
...
+                # not an entered side — skip it so doubling a position is not
+                # misread as a duplicate exit. A close whose side cannot be
+                # resolved statically is judged against every entered side
+                # (forbid if any is engine-owned).
+                closed_side = _submit_order_close_side(node)
+                if closed_side is not None:
+                    if closed_side not in entered_sides:
+                        continue
+                    sides_to_check = {closed_side}
+                else:
```
brandonkindred: **[INFO] logic** — When `_submit_order_close_side` returns `None` (side cannot be resolved statically), the duplicate-exit check tests all entered sides. If any entered side is engine-owned, the close is flagged. This conservative approach may produce false positives for strategies that dynamically select an uncovered side at runtime. The design is intentional but could reject valid strategies.

**Suggested fix:** Add a comment in the docstring or inline explaining the trade-off, and consider providing an escape hatch (e.g., a decorator or comment pragma) for strategy authors to assert that a dynamic close only targets uncovered sides.

backend/agents/investment_team/strategy_lab/prompts/code_synthesis_system.md
```diff
@@ -71,10 +71,14 @@ class MyStrategy(Strategy):
             # <FILL IN entry decision per spec.entry_rules>
             pass
 
-        # ── EXIT ──────────────────────────────────────────
-        else:
-            # <FILL IN exit decision per spec.exit_rules>
-            pass
+        # ── ENGINE OWNS EXITS (do NOT submit close orders) ──
```
brandonkindred, line 75: **[INFO] standards** — The prompt instructs the LLM not to submit close orders, but does not explicitly state that indicators referenced only by engine-owned exit rules do not need to be read in `on_bar`. While the conformance gate will not enforce them, the LLM may still generate unnecessary indicator reads, adding complexity.

**Suggested fix:** Add a sentence after line 82: 'Likewise, indicators that are only used by `spec.exit_rules` (e.g., an RSI for a signal-exit) are computed by the engine; you do not need to read them in `on_bar`.'

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
brandonkindred: Documented in `daad812`. `_submit_order_close_side` inspects only the keyword `side=` form — matching `_submit_order_closes_position` (keyword `qty=`) and the synthesis prompt's mandated keyword-argument call shape. A positional `side` also escapes `_submit_order_closes_position`, so such a call never reaches this helper as a close in the first place. I documented the keyword-only contract rather than adding positional parsing, whose param order isn't guaranteed stable.

---

backend/agents/investment_team/strategy_lab/quality_gates/predicate_conformance.py
brandonkindred: Fixed in `daad812` — `entered_sides` now filters out `None` in both the duplicate-exit check and the predicate gate. In practice `EntryRule.side` is a validated `Literal["long", "short"]` (default `long`) so it's never `None`; the filter is belt-and-suspenders against a future schema change.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
brandonkindred: Documented in `daad812` — expanded the inline comment to spell out the conservative trade-off: a dynamic close that only ever targets an uncovered side at runtime may be a false positive, and the gate prefers a refinement round over silently letting a duplicate engine exit through. I didn't add a pragma escape hatch — it would be the only such mechanism in these gates, and the false-positive surface is narrow (a `qty=position.qty` close with a non-literal `side=`).

---

backend/agents/investment_team/strategy_lab/prompts/code_synthesis_system.md
brandonkindred: Added in `daad812` — the "Exits are engine-owned" section now states that indicators referenced only by `spec.exit_rules` (e.g. an RSI used solely by a signal-exit) are computed by the engine and need not be read in `on_bar`; read only the indicators your entry logic uses. (This mirrors the gate change from the first commit, where `_collect_required_indicators` stopped requiring exit-only indicator reads on the custom-code path.)

---

brandonkindred:

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -807,61 +865,160 @@ def _check_entry_coverage(self, cctx: _ConformanceCtx) -> Iterable[QualityGateRe
...
+            for r in (getattr(spec, "entry_rules", []) or [])
+            if isinstance(r, EntryRule) and r.side is not None
+        }
+
+        n_closes = 0
+        for method in _iter_strategy_methods(cctx.cls):
+            if method.name not in cctx.reachable:
+                continue
+            for node in _iter_method_body_nodes(method):
+                if not _is_submit_order_call(node) or not _submit_order_closes_position(node):
```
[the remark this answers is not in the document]
brandonkindred: Fixed in `ba9a8a8`. The duplicate-exit check now resolves position aliases and wrapped expressions: it builds the alias set with `_collect_position_aliases` (which picks up `current = ctx.position(...)`) and detects the close via `_expr_references_position_qty` (which walks the `qty=` expression tree, so `current.qty`, `abs(position.qty)`, `pos.qty * 1`, etc. all match) — the same `code_safety` machinery you pointed to, wrapped in a new `_submit_order_qty_references_position` helper. Added `test_no_duplicate_engine_exit_detects_aliased_position_close` and `..._detects_wrapped_position_qty_close`.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -374,6 +389,50 @@ def _submit_order_closes_position(call: ast.Call) -> bool:
...
+    """
+    for kw in call.keywords:
+        if kw.arg != "side":
+            continue
+        v = kw.value
+        token: Optional[str] = None
+        if isinstance(v, ast.Constant) and isinstance(v.value, str):
+            token = v.value.lower()
+        elif isinstance(v, ast.Attribute):
+            token = v.attr.lower()
```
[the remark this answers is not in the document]
brandonkindred: Fixed in `ba9a8a8`. `_submit_order_close_side` now delegates to `code_safety._extract_side_literal`, which only accepts an `OrderSide`-rooted attribute (plus `OrderSide(...)` calls and `"LONG"`/`"SHORT"` strings). A non-`OrderSide` enum like `CustomSide.SHORT` now resolves to `None` — treated as an unresolvable side (judged conservatively against all entered sides) rather than misread as a short order. This keeps side inference identical to `code_safety`'s entry-side detection. Added `test_no_duplicate_engine_exit_resolves_orderside_attribute` covering the `OrderSide.SHORT` attribute form.

---

brandonkindred:

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -807,61 +880,168 @@ def _check_entry_coverage(self, cctx: _ConformanceCtx) -> Iterable[QualityGateRe
...
+                # misread as a duplicate exit. A close whose side cannot be
+                # resolved statically is judged against every entered side
+                # (forbid if any is engine-owned). That is deliberately
+                # conservative: a dynamic close that only ever targets an
+                # uncovered side at runtime may be flagged as a false
+                # positive — the gate prefers a refinement round over
+                # silently letting a duplicate engine exit through.
+                closed_side = _submit_order_close_side(node)
+                if closed_side is not None:
+                    if closed_side not in entered_sides:
```
brandonkindred: **[MEDIUM] logic** — The scale-in guard in `_check_no_duplicate_engine_exit` only skips a position-qty order when `closed_side not in entered_sides`. For a spec that enters both long and short, a scale-in order (e.g., `side=LONG, qty=position.qty` to add to an existing long) would have `closed_side="short"`, which is in `entered_sides`, so the guard does not skip it. If the engine covers the short side, the order is incorrectly flagged as a duplicate engine exit, even though it is not closing any position. This can block valid strategies that use `qty=position.qty` for scaling into a position when the spec supports both sides.

**Suggested fix:** Consider refining the guard to also check whether the order side matches the current position side (e.g., by comparing the order side literal to the position's side, if statically known). Since static analysis cannot determine runtime position side, an alternative is to document this limitation explicitly in the gate's docstring and error message, and possibly allow a suppression mechanism for multi-side specs. At minimum, update the docstring to acknowledge this false-positive scenario.

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -807,61 +880,168 @@ def _check_entry_coverage(self, cctx: _ConformanceCtx) -> Iterable[QualityGateRe
         return ()
 
     # ------------------------------------------------------------------
-    # Check 4 — signal-exit predicate coverage
+    # Check 4 — signal-exit rules are engine-owned
     # ------------------------------------------------------------------
     def _check_signal_exit_coverage(self, cctx: _ConformanceCtx) -> Iterable[QualityGateResult]:
```
brandonkindred: **[LOW] naming** — The method `_check_signal_exit_coverage` now only emits an informational message and never a critical. The name implies it checks coverage, which is misleading and may confuse maintainers.

**Suggested fix:** Rename to `_note_signal_exit_engine_ownership` or `_inform_signal_exit_engine_owned` to reflect its current purpose.

backend/agents/investment_team/strategy_lab/prompts/alignment_propose_fix.md
```diff
@@ -61,7 +61,18 @@ A JSON object with these fields:
...
   hardcode integer share counts.
 - Do not implement bar-counting "time stop" exits (e.g. ``bars_held``,
-  ``hold_count``, ``if counter >= N: close``). Exits must use price,
-  P&L, or signal-based logic only.
+  ``hold_count``, ``if counter >= N: close``); they are rejected by the
+  conformance gate.
+- Exits are engine-owned. The engine enforces every ``spec.exit_rules``
+  entry (stop-loss / take-profit / signal-exit) for the side(s) it
+  applies to and stamps ``engine_exit:<kind>`` attribution. When a
+  finding cites a ``signal_exit`` or ``take_profit`` divergence, the fix
```
brandonkindred: **[MEDIUM] spec-compliance** — The instruction tells the repair agent to remove manual closes only for `signal_exit` and `take_profit` divergences. However, the engine also owns stop-loss exits, and the alignment gate may produce stop-loss divergence findings. If it does, the repair agent might not know to remove the manual close, leading to an incorrect fix.

**Suggested fix:** Change the text to "When a finding cites a ``signal_exit``, ``take_profit``, or ``stop_loss`` divergence" or generalize to "any exit rule divergence" to cover all engine-owned exit types.

backend/agents/investment_team/strategy_lab/prompts/code_synthesis_system.md
```diff
@@ -62,7 +62,7 @@ class MyStrategy(Strategy):
         # ── COMPUTE SIGNALS ───────────────────────────────
         # Read each spec indicator via ``ctx.indicator(name, **params)``.
```
brandonkindred: **[LOW] standards** — The template comment says "Read each spec indicator", which could be interpreted as reading all indicators declared in the spec, including those used only by exit rules. The later section clarifies that exit-only indicators are not needed, but the template comment may lead the LLM to read them anyway, potentially causing unnecessary code.

**Suggested fix:** Change the comment to "Read each indicator your entry logic needs (see spec.entry_rules)" or "Read entry-rule indicators via ``ctx.indicator(...)``" to align with the engine-owned-exits contract.

backend/agents/investment_team/strategy_lab/quality_gates/predicate_conformance.py
```diff
@@ -313,6 +325,18 @@ def check(
...
             fixtures = generate_conformance_fixtures(spec, compiled_code=code)
+            # Signal exits are engine-owned: when spec.exit_rules cover every
+            # entered side, the engine enforces the SignalExitRule via
+            # _EngineExitDispatcher and conforming custom code submits no manual
+            # close. The shadow context only sees strategy-submitted orders, so
+            # keeping those fixtures would demand a close the conformance
+            # contract now forbids. Drop them; entry fixtures still run
+            # (entries remain inline for custom code).
+            entered_sides = {
+                r.side for r in entry_rules if isinstance(r, _EntryRule) and r.side is not None
```
brandonkindred, line 336: **[INFO] standards** — The name `_EntryRule` is used but its import is not visible in the provided snippet (imports shown at lines 51-56 do not include it). Ensure it is imported (likely from `..spec_dsl`) in the full file to avoid NameError.

**Suggested fix:** Add `from ..spec_dsl import EntryRule as _EntryRule` to the imports if not already present in the omitted lines.

brandonkindred (commented):
The pull request correctly implements the engine-owned exits contract for custom-code strategies. The review covered changes to prompts, the code conformance gate, the predicate conformance gate, and associated tests. The implementation is well-structured, thoroughly documented with Design by Contract, and supported by comprehensive tests covering edge cases. The code is production-ready, with a few minor issues noted.

Findings, ordered by severity:
- [medium] In code_conformance.py, the scale-in guard in `_check_no_duplicate_engine_exit` may produce a false positive for multi-side specs when a scale-in order uses `qty=position.qty` and the engine covers the opposite side. This could block valid strategies. Refining the guard or documenting the limitation is suggested.
- [medium] In alignment_propose_fix.md, the repair agent instruction to remove manual closes only mentions `signal_exit` and `take_profit` divergences, omitting `stop_loss`. If the alignment gate produces a stop-loss divergence, the agent might not remove the manual close, leading to an incorrect fix. The text should be generalized to cover all engine-owned exit types.
- [low] The method `_check_signal_exit_coverage` in code_conformance.py now only emits informational messages, making its name misleading. Renaming to reflect its current purpose is recommended.
- [low] A template comment in code_synthesis_system.md could be misinterpreted to read all spec indicators, including exit-only ones, potentially causing unnecessary code. Clarifying the comment to focus on entry-rule indicators would align with the engine-owned-exits contract.
- [info] In predicate_conformance.py, the name `_EntryRule` is used but its import is not visible in the provided snippet; ensure it is imported to avoid a NameError.

**Spec compliance:** The submission fully meets the acceptance criteria. The prompts remove manual exit guidance, the alignment repair now removes duplicates, the code conformance gate retires the signal-exit requirement and adds a duplicate-exit prohibition, and the predicate conformance gate skips signal-exit fixtures when the engine covers sides. Indicator requirements are adjusted accordingly. Tests confirm the behavior. Minor gaps exist: the repair agent prompt does not explicitly cover stop-loss divergences, and the duplicate-exit check may have a false positive for scale-in orders in multi-side specs. These do not undermine the core contract, and the implementation is compliant overall.

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -807,61 +880,168 @@ def _check_entry_coverage(self, cctx: _ConformanceCtx) -> Iterable[QualityGateRe
...
+
+        n_closes = 0
+        for method in _iter_strategy_methods(cctx.cls):
+            if method.name not in cctx.reachable:
+                continue
+            for node in _iter_method_body_nodes(method):
+                if not _is_submit_order_call(node):
+                    continue
+                if not _submit_order_qty_references_position(node, position_names):
+                    continue
```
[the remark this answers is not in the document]
brandonkindred: Fixed in `f888055`. Added `_submit_order_attached_bracket_side`: an entry passing a non-`None` `attached_stop_loss` / `attached_take_profit` is now treated as a manual exit for the order's OWN side (the `bracket_sl` / `bracket_tp` children close the position the entry opens, ahead of `_EngineExitDispatcher`), and flagged when that side is engine-owned. Test: `test_no_duplicate_engine_exit_detects_attached_bracket_leg`.

---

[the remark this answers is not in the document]
brandonkindred: Fixed the alias case in `f888055` — `close_qty = pos.qty; submit_order(qty=close_qty)` is now caught via a one-hop qty-alias set (`_collect_qty_aliases`). Test: `test_no_duplicate_engine_exit_detects_qty_alias_close`.

For the harder "opposite-side calculated quantity that never references `position.qty`" shape: telling it apart from a same-side entry requires the runtime position side, which static analysis doesn't have (a `side=SHORT` order is a short *entry* or a long *close* depending on whether a position is open). I documented this — together with the two-sided same-side-scale-in ambiguity — as a known static limit in the gate docstring, with the **runtime trade-alignment gate** as the authoritative backstop: it verifies `engine_exit:<kind>` attribution per executed trade and catches any attribution-stripping close the static pass misses. I can pursue a fuller opposite-side-in-position-open-branch analysis if you'd prefer it caught statically, but it carries real false-positive risk (it would flag legitimate short entries) — happy to do it if you want that trade-off.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
brandonkindred: Documented in `f888055`. The gate docstring now states the limitation explicitly: in a spec that enters BOTH sides, a full-size same-side scale-in cannot be told apart from a cross-side close without the runtime position side, so it may be flagged (false positive) or a cross-side close waved through — and the runtime trade-alignment gate reconciles both. I kept the static gate's conservative behavior rather than adding a suppression pragma (it would be the only such mechanism in these gates); let me know if you'd prefer the pragma.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
brandonkindred: Done in `f888055` — renamed to `_note_signal_exit_engine_ownership` (and updated the call site in `check()`), since it now only emits the informational engine-ownership row and never a critical.

---

backend/agents/investment_team/strategy_lab/prompts/alignment_propose_fix.md
brandonkindred: Fixed in `f888055` — the repair constraint now cites a `signal_exit`, `take_profit`, or `stop_loss` divergence, covering all three engine-owned exit kinds.

---

backend/agents/investment_team/strategy_lab/prompts/code_synthesis_system.md
brandonkindred: Fixed in `f888055` — the COMPUTE SIGNALS comment now reads "Read each indicator your ENTRY logic needs (see spec.entry_rules) via `ctx.indicator(name, **params)`", so the LLM is steered away from reading engine-owned exit-only indicators.

---

backend/agents/investment_team/strategy_lab/quality_gates/predicate_conformance.py
brandonkindred: No change needed — `_EntryRule` is imported: `from ..spec_dsl import EntryRule as _EntryRule` at line 43 of `predicate_conformance.py`, above the range shown in the snippet. The full suite (which exercises this path) passes, confirming there's no `NameError`.

---

brandonkindred:

backend/agents/investment_team/strategy_lab/prompts/code_synthesis_system.md
```diff
@@ -173,13 +178,27 @@ Do NOT import: `pandas`, `numpy`, `os`, `sys`, `subprocess`, `socket`, `http`, `
...
-Exits must be based on price (stop-loss), P&L (take-profit), or signal
-reversal (signal_exit) only.
+## Exits are engine-owned
+
+The engine enforces every exit in `spec.exit_rules` — stop-loss,
+take-profit, and signal-exit — for the position side(s) each rule
+applies to, and stamps `engine_exit:<kind>` on the close. Do NOT author
+a position-closing order for a side the engine covers: a manual close
+fills ahead of the engine, strips that attribution, and fails the
+trade-alignment gate. The one exception is a position side that **no**
```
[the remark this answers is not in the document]
brandonkindred: Addressed the actionable part in `a7a8e6c`: the `zero_trade_repair` prompt's `ENTRY_WITH_NO_EXIT` guidance now says to fix the non-firing exit at the **spec level** (loosen thresholds / add a `signal_exit` to `spec.exit_rules` / widen bands — all engine-enforced) and explicitly NOT to add a manual code close. That keeps a valid repair path: the zero-trade repair edits the spec (it already lists `exit_rules`/`sizing`/`risk_limits` as editable), so a looser engine exit is allowed and re-enforced by the engine. The broader "route `ENTRY_WITH_NO_EXIT` to a dedicated spec-refinement stage" is an orchestrator routing change beyond this gate PR — happy to open a follow-up issue for it.

---

backend/agents/investment_team/strategy_lab/prompts/code_synthesis_system.md
```diff
@@ -71,10 +72,14 @@ class MyStrategy(Strategy):
...
-        else:
-            # <FILL IN exit decision per spec.exit_rules>
-            pass
+        # ── ENGINE OWNS EXITS (do NOT submit close orders) ──
+        # The engine enforces every spec.exit_rules entry (stop-loss,
+        # take-profit, signal-exit) for the side(s) it applies to and
+        # stamps engine_exit:<kind> on the close. Do NOT submit a closing
+        # order for a side the engine covers — a manual close fills first,
+        # strips that attribution, and fails the trade-alignment gate. The
+        # only close you author is for a position side NO exit rule covers
```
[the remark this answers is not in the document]
brandonkindred: Fixed in `a7a8e6c` — `refinement_system.md` line 93 no longer instructs an opposite-side `qty==position.qty` close. It now carries the engine-owned-exit contract: submit entries; the engine enforces every `spec.exit_rules` entry and a manual close for a covered side is rejected by the conformance gate (only close a side no exit rule covers).

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -374,6 +394,118 @@ def _submit_order_closes_position(call: ast.Call) -> bool:
...
+    """
+    names: set[str] = set()
+    for node in ast.walk(cls):
+        if not isinstance(node, ast.Assign):
+            continue
+        if not _expr_references_position_qty(node.value, position_names):
+            continue
+        for target in node.targets:
+            if isinstance(target, ast.Name):
+                names.add(target.id)
```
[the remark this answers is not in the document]
brandonkindred: Agreed — fixed in `a7a8e6c`. I removed the class-wide `_collect_qty_aliases`: it was flow-insensitive and bound any name assigned `= pos.qty` anywhere in the class, so a legitimately-computed entry qty sharing that name was flagged as a close (the hard-critical false positive you describe). Detection is now per-call (`_expr_is_position_qty`): direct `pos.qty`/position-alias, wrapped expressions, and direct `ctx.position(...).qty`. Name-/attribute-bound qty aliases are a documented static limit, caught by the runtime trade-alignment gate. Regression test: `test_no_duplicate_engine_exit_no_false_positive_on_computed_qty_entry`.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -807,61 +938,182 @@ def _check_entry_coverage(self, cctx: _ConformanceCtx) -> Iterable[QualityGateRe
...
+                continue
+            for node in _iter_method_body_nodes(method):
+                if not _is_submit_order_call(node):
+                    continue
+                closed_sides: set[str] = set()
+                if _submit_order_qty_references_position(node, position_names, qty_alias_names):
+                    side = _submit_order_close_side(node)
+                    # Resolved side → the one opposite side it retires;
+                    # unresolvable side → fall back to every entered side.
+                    closed_sides |= {side} if side is not None else set(entered_sides)
```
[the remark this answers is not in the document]
brandonkindred: Partially fixed in `a7a8e6c` — `qty=ctx.position(bar.symbol).qty` (direct receiver) is now detected via `_expr_is_position_qty` (test `..._detects_direct_ctx_position_qty_close`). The attribute-alias form (`self.current = pos; qty=self.current.qty`) is the same class as name-bound qty aliases, which I deliberately left to the runtime trade-alignment gate after the class-wide alias collection caused false positives (your line-478 thread). Resolving attribute aliases statically reintroduces that FP surface, so I kept it runtime-caught and documented the limit in the gate docstring.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -374,6 +394,118 @@ def _submit_order_closes_position(call: ast.Call) -> bool:
...
+    has_bracket = any(
+        kw.arg in ("attached_stop_loss", "attached_take_profit")
+        and not (isinstance(kw.value, ast.Constant) and kw.value.value is None)
+        for kw in call.keywords
+    )
+    if not has_bracket:
+        return None
+    for kw in call.keywords:
+        if kw.arg == "side":
+            return _extract_side_literal(kw.value)
```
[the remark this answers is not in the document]
brandonkindred: Fixed in `a7a8e6c` — `_submit_order_attached_bracket` now returns `(has_bracket, side)`, and when the bracket's order side is dynamic (`_extract_side_literal` returns `None`) the loop falls back to every entered side, exactly as it does for an explicit close with an unresolvable side.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -374,6 +394,118 @@ def _submit_order_closes_position(call: ast.Call) -> bool:
...
+    Post: every ``<name> = <expr>`` assignment anywhere in ``cls`` whose RHS
+    satisfies :func:`_expr_references_position_qty` contributes ``<name>``,
+    so the close detector sees ``close_qty = pos.qty; submit_order(qty=close_qty)``.
+    Only one indirection hop is resolved (a name bound directly to a
+    position-qty expression); deeper chains fall to the runtime gate.
+    """
+    names: set[str] = set()
+    for node in ast.walk(cls):
+        if not isinstance(node, ast.Assign):
+            continue
```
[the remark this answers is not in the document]
brandonkindred: With the class-wide qty-alias collection removed (your line-478 thread), there's no longer an `ast.Assign`-only collector to extend — name aliases via either `Assign` or `AnnAssign` are now uniformly a documented static limit caught by the runtime gate. So `close_qty: float = pos.qty; qty=close_qty` is handled the same as the plain-assignment form: not statically flagged, runtime-caught. (Adding `AnnAssign` to a class-wide collector would just reintroduce the false-positive surface that motivated removing it.)

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -374,6 +394,118 @@ def _submit_order_closes_position(call: ast.Call) -> bool:
...
+    order opens, before ``_EngineExitDispatcher`` stamps
+    ``engine_exit:<kind>``. Returns ``None`` when no non-``None`` bracket leg
+    is attached or the order side is not a resolvable ``OrderSide`` literal.
+    Unlike an explicit close the bracket retires the order's OWN side, so the
+    side is NOT inverted.
+    """
+    has_bracket = any(
+        kw.arg in ("attached_stop_loss", "attached_take_profit")
+        and not (isinstance(kw.value, ast.Constant) and kw.value.value is None)
+        for kw in call.keywords
```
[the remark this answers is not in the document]
brandonkindred: Fixed in `a7a8e6c` — `_attachment_is_active` now treats a bracket attachment as active only when its value is demonstrably non-`None` (a non-`None` constant or a computed expression). A bare name/attribute (`stop = None; attached_stop_loss=stop`) or a literal `None` is treated as inactive, so it's no longer a false-positive critical; a genuinely-active bracket built from a variable is caught by the runtime gate. Test: `test_no_duplicate_engine_exit_ignores_none_attachment`.

---

brandonkindred:
@codex review

brandonkindred:

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -807,61 +951,189 @@ def _check_entry_coverage(self, cctx: _ConformanceCtx) -> Iterable[QualityGateRe
...
+                continue
+            for node in _iter_method_body_nodes(method):
+                if not _is_submit_order_call(node):
+                    continue
+                closed_sides: set[str] = set()
+                if _submit_order_qty_references_position(node, position_names):
+                    side = _submit_order_close_side(node)
+                    # Resolved side → the one opposite side it retires;
+                    # unresolvable side → fall back to every entered side.
+                    closed_sides |= {side} if side is not None else set(entered_sides)
```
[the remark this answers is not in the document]
brandonkindred: Fixed in `3969e69` — good catch, the unresolved-side fallback was over-flagging. A close (or attached bracket) whose `side=` can't be resolved statically is now a hard critical **only when the engine covers every entered side** (then any close hits a covered side — there's no legitimate uncovered side to retire). Under partial coverage it's deferred to the runtime trade-alignment gate, preserving the partial-coverage carve-out. Resolved-side closes are unchanged (judged on their specific side).

Your exact scenario — long+short entries, long-only `trailing_high`, `close_side = OrderSide.LONG; submit_order(side=close_side, qty=pos.qty)` closing the uncovered short — is now allowed (`test_no_duplicate_engine_exit_dynamic_side_close_allowed_under_partial_coverage`); the all-covered case still flags (`..._flagged_when_all_covered`).

---

brandonkindred:

backend/agents/investment_team/strategy_lab/prompts/zero_trade_repair_system.md
```diff
@@ -66,10 +66,14 @@ together tell you where in the lifecycle execution stopped.
...
   reference an unset attribute, or only fire on conditions that never
-  occur in the test window. Loosen the exit predicate thresholds,
-  add a signal-based fallback exit (e.g. a reversal or momentum-
-  exhaustion condition), or widen the stop-loss / take-profit bands
-  so trades close within the test window. Do NOT add bar-counting
+  occur in the test window. Fix this at the SPEC level — exits are
+  engine-owned: loosen the exit predicate thresholds, add a
+  signal-based fallback to `spec.exit_rules` (e.g. a reversal or
+  momentum-exhaustion `signal_exit`), or widen the stop-loss /
+  take-profit bands so the engine closes trades within the test window.
```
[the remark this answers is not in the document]
brandonkindred: Fixed the prompt in `fb37a01` — you're right, this was a dead-end. Since `exit_rules` spec edits are dropped (only `risk_limits` is honoured) and this PR forbids manual code closes, the prior "fix at the spec level" guidance had no actionable repair. The prompt now states `ENTRY_WITH_NO_EXIT` is **not code-repairable** here and tells the agent to report the category with evidence and propose no rewrite, so the run falls through to redesign rather than emitting discarded updates. Per the owner's direction I'm also adding the orchestrator routing for this category (route to spec redesign) as the follow-up — leaving this open until that lands.

---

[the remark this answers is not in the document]
brandonkindred: Same fix as the sibling thread — `fb37a01`. The `ENTRY_WITH_NO_EXIT` guidance no longer instructs `spec.exit_rules` edits (they're discarded). Orchestrator routing to redesign is the follow-up; leaving open until that lands.

---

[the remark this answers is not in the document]
brandonkindred: Fixed in `fb37a01` — good catch. Widening the bands moves the trigger farther from entry and makes an exit *less* likely, worsening `ENTRY_WITH_NO_EXIT`. The reverted guidance removes the band-change instruction entirely (and explicitly notes widening is counterproductive); since `exit_rules` edits are dropped anyway, the agent doesn't touch the bands at all.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -807,61 +951,208 @@ def _check_entry_coverage(self, cctx: _ConformanceCtx) -> Iterable[QualityGateRe
...
+                if not _is_submit_order_call(node):
+                    continue
+                resolved_sides: set[str] = set()
+                unresolved = False
+                if _submit_order_qty_references_position(node, position_names):
+                    side = _submit_order_close_side(node)
+                    if side is not None:
+                        resolved_sides.add(side)
+                    else:
+                        unresolved = True
```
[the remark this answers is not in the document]
brandonkindred: You're correct on both points — a `**kwargs`-spread close evades the static check, **and** the runtime alignment gate is not a complete backstop (it accepts a manual close that complies with the spec's exit bounds; only the `engine_exit` label is lost). Rather than keep chasing static evasions (which conflict with the false-positive concerns elsewhere in this same review), the owner and I decided to move exit-ownership enforcement to the **engine**: it will reconcile attribution — stamping `engine_exit:<kind>` when a structured rule fires (within bounds) on the bar a manual close fills. That makes the runtime path complete (compliant closes get correct attribution; non-compliant ones are still flagged as divergences) and renders these static FN gaps moot. Implementing that next; the static gate stays as a best-effort fast-fail. Leaving open until the engine reconciliation lands.

---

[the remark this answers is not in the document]
brandonkindred: Agreed, and your "not reliably caught later" point is the key one — the runtime alignment gate passes a spec-compliant manual close without `engine_exit` attribution. We're addressing this at the engine level (attribution reconciliation: stamp `engine_exit:<kind>` when a structured rule fires within bounds on the close bar), which makes the name-alias / dict-spread / direct-call evasions moot for compliant closes and keeps non-compliant ones flagged. The class-wide static alias resolution was removed earlier precisely because it caused false positives, so engine-side reconciliation is the right layer. Implementing next; leaving open until it lands.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -374,6 +395,130 @@ def _submit_order_closes_position(call: ast.Call) -> bool:
...
+    (``stop = None; attached_stop_loss=stop``). An engine that treats a
+    falsy attachment as "no bracket" creates no bracket child, so a falsy
+    or ambiguous attachment is treated as inactive to avoid a false-positive
+    hard critical; a genuinely-active bracket built from a variable is caught
+    by the runtime trade-alignment gate instead.
+    """
+    if isinstance(node, ast.Constant):
+        return bool(node.value)
+    if isinstance(node, (ast.Name, ast.Attribute)):
+        return False
```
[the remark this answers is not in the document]
brandonkindred: Right — a named bracket (`stop = StopAttachment(...); attached_stop_loss=stop`) is treated as inactive to avoid the false positive on `stop = None`, so it's a static FN; and the bracket close can satisfy the alignment thresholds without `engine_exit` attribution. Same disposition as the other FN threads: the engine-side attribution reconciliation (stamp `engine_exit:<kind>` when the structured rule fires within bounds on the close/bracket bar) closes this at the right layer, where the runtime position/price is known. Implementing next; leaving open until it lands.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -807,61 +951,208 @@ def _check_entry_coverage(self, cctx: _ConformanceCtx) -> Iterable[QualityGateRe
...
+                continue
+            for node in _iter_method_body_nodes(method):
+                if not _is_submit_order_call(node):
+                    continue
+                resolved_sides: set[str] = set()
+                unresolved = False
+                if _submit_order_qty_references_position(node, position_names):
+                    side = _submit_order_close_side(node)
+                    if side is not None:
+                        resolved_sides.add(side)
```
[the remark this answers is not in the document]
brandonkindred: Agreed — `qty=position.qty / 2` is a partial reduction, not a full close, and can't overwrite the final engine-exit attribution, so flagging it is a false positive. With the engine-reconciliation direction the static gate becomes a secondary fast-fail, but I'll tighten it to only flag a **full**-position close (exact `position.qty` / `abs(position.qty)`, not an arithmetic fraction) as part of the same change. Leaving open until that lands.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -807,61 +951,208 @@ def _check_entry_coverage(self, cctx: _ConformanceCtx) -> Iterable[QualityGateRe
...
+                    # Bracket closes the order's OWN side.
+                    if bracket_side is not None:
+                        resolved_sides.add(bracket_side)
+                    else:
+                        unresolved = True
+                # A real close retires a side the spec actually enters; the
+                # intersection also drops a same-side scale-in whose inferred
+                # opposite side the spec never enters.
+                resolved_sides &= entered_sides
+                flagged = any(_engine_exits_cover_sides(spec, {s}) for s in resolved_sides)
```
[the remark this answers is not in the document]
brandonkindred: Agreed — this is the documented two-sided same-side-scale-in ambiguity (`side=LONG, qty=pos.qty` to double a long is inferred as a short close, and `short` being entered + covered emits a critical). It's a genuine false positive that static analysis can't disambiguate without the runtime position side. With the engine-reconciliation direction this static check becomes secondary; I'll tighten the FP (and the partial-reduction one) as part of that change. Leaving open until it lands.

---

backend/agents/investment_team/strategy_lab/prompts/zero_trade_repair_system.md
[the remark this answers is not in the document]
brandonkindred: Tracked in #874 (orchestrator routing of `ENTRY_WITH_NO_EXIT` to redesign). Prompt already corrected in this PR; resolving here.

---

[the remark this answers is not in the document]
brandonkindred: Tracked in #874. Resolving here.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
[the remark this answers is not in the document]
brandonkindred: Tracked in #873 (engine-side attribution reconciliation). Resolving here.

---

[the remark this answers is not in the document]
brandonkindred: Tracked in #873. Resolving here.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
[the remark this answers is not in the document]
brandonkindred: Tracked in #873. Resolving here.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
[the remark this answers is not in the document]
brandonkindred: Tracked in #873 (static-gate FP cleanup, enabled by the engine reconciliation). Resolving here.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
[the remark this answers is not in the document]
brandonkindred: Tracked in #873 (static-gate FP cleanup). Resolving here.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -807,61 +951,208 @@ def _check_entry_coverage(self, cctx: _ConformanceCtx) -> Iterable[QualityGateRe
...
+                # A dynamic (unresolved) side is a hard critical only when the
+                # engine covers every entered side — otherwise it may be the
+                # legitimate close of an engine-uncovered side.
+                if unresolved and all_sides_covered:
+                    flagged = True
+                if flagged:
+                    n_closes += 1
+        if n_closes == 0:
+            return ()
+        return (
```
brandonkindred, line 1152: **[MEDIUM] standards** — The critical result returned by `_check_no_duplicate_engine_exit` does not include the names of the methods or the source locations where the duplicate closes were found. The message only states the count of manual closes, which may make it harder for downstream repair agents (or developers) to quickly locate the offending `submit_order` calls in the strategy code.

**Suggested fix:** Enhance the critical details to include the method names (e.g., `on_bar`) or AST line numbers where the duplicate closes were detected. For example, collect a list of method names during the loop and append them to the `details` field of the `QualityGateResult`, e.g.: `f"Custom-code strategy submits {n_closes} manual position-closing order(s) in methods: {', '.join(method_names)} — ..."`. This provides actionable location information without requiring a full AST dump.

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
```diff
@@ -807,61 +951,208 @@ def _check_entry_coverage(self, cctx: _ConformanceCtx) -> Iterable[QualityGateRe
...
-                    f"Spec declares {len(signal_exits)} signal-exit rule(s) "
-                    f"but only {len(exit_branches)} exit branch(es) were found "
-                    "reachable from on_bar."
-                ),
-            )
-        return ()
+    # ------------------------------------------------------------------
+    # Check 4b — no manual close duplicating an engine-owned exit
+    # ------------------------------------------------------------------
+    def _check_no_duplicate_engine_exit(self, cctx: _ConformanceCtx) -> Iterable[QualityGateResult]:
```
brandonkindred, line 1044: **[LOW] standards** — The method `_check_no_duplicate_engine_exit` is relatively long and handles several distinct responsibilities: resolving position aliases, iterating reachable methods, detecting explicit closes and attached brackets, resolving sides, and evaluating coverage. While the logic is sound and well-documented, extracting some of these concerns into smaller private helper methods would improve readability and make the method easier to maintain.

**Suggested fix:** Consider refactoring by extracting the per-node close-detection and flagging logic into a separate helper (e.g., `_is_duplicate_close(node, position_names, entered_sides, spec) -> bool`), and keeping the main method focused on orchestration: collecting aliases, iterating methods, and aggregating results. This would reduce the method's length and clarify the separation between detection and policy evaluation.

brandonkindred (commented):
The PR correctly implements the engine-owned exits contract for custom-code strategies. Prompts are updated to forbid manual closes for engine-covered sides and to instruct removal of duplicate closes. The code conformance gate adds a well-designed, coverage-gated hard critical that detects manual closes and attached brackets duplicating engine-owned exits, while properly allowing legitimate closes for uncovered sides. The predicate conformance gate correctly skips signal-exit fixtures when the engine covers all entered sides. Tests are comprehensive and cover the main scenarios, edge cases, and documented static limits. Code quality is high with thorough Design by Contract documentation. Minor improvements could be made to the diagnostic detail of the critical message and to method length, but these do not block approval.

**Spec compliance:** All acceptance criteria are met: (1) code_synthesis_system.md removes the manual exit branch and declares engine ownership; (2) alignment_propose_fix.md instructs removal of duplicate closes; (3) code_conformance.py retires the manual signal-exit requirement, adds the duplicate-exit critical, and stops requiring exit-only indicators; (4) predicate_conformance.py skips signal-exit fixtures when engine covers entered sides; (5) tests cover the new behavior, including the inverted requirement, partial coverage, brackets, aliases, kwargs-spread carve-out, and dynamic-side resolution. The implementation aligns with the PR description and the overall project specification.

- `backend/agents/investment_team/tests/test_code_conformance_gate.py` — **[INFO] testing** — The test file uses several helper functions (`_spec`, `_happy_spec`, `_HAPPY_CODE`, `_rsi_signal_exit`, `_sma_cross_entry`, `_critical_details`) that are not defined in the provided code chunk. While they are likely defined elsewhere in the same file or in a shared conftest, their absence from the review chunk makes it impossible to verify their correctness and alignment with the current spec models.

**Suggested fix:** Ensure these test helpers are defined in a visible location (e.g., at the top of the test file or in a `conftest.py`) and are kept up-to-date with any changes to the `StrategySpec` model. This prevents test brittleness and makes the test suite self-contained for future reviewers.

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
brandonkindred: Done in `aa3ff19` — the critical now lists each offending close's location as `<method>:L<line>` (e.g. `on_bar:L18`), so the refinement agent / developer can jump straight to the call instead of only seeing a count. Added a test asserting the `on_bar:L<line>` location is present in the message.

---

backend/agents/investment_team/strategy_lab/quality_gates/code_conformance.py
brandonkindred: Done in `aa3ff19` — extracted the per-node close/bracket detection + coverage policy into a module-level `_node_is_duplicate_close(node, *, position_names, entered_sides, all_sides_covered, spec) -> bool`. `_check_no_duplicate_engine_exit` is now focused on orchestration (resolve aliases, compute coverage once, iterate reachable methods, aggregate locations). Behaviour is unchanged — full suite green.

---

PR #870, opened 2026-06-12
Require context-specific answer options for coding team clarification questions

brandonkindred:
## Summary

- **Root cause**: The Tech Lead and Senior SWE agent prompts marked `options` as optional in their `open_questions` schema. The LLM never generated them, so every question fell through to the `DEFAULT_CLARIFICATION_OPTIONS` fallback — a hardcoded yes/no/not-sure set that is inappropriate for open-ended questions like "What API fields are needed?" or "Who owns the legal disclaimer copy?"
- **Prompt fix**: Both agent prompts now explicitly require ≥2 context-specific answer options per question, modeled after the planning team's approach. Options must represent actual implementation choices (e.g., auth strategies, team names, UI patterns), not generic yes/no. An `is_default` flag marks the safest/most-common choice.
- **Fallback fix**: `DEFAULT_CLARIFICATION_OPTIONS` in `coding_team/hitl.py` and `software_engineering_team/orchestrator.py` is now an empty list. If the LLM unexpectedly omits options, the question renders with only the always-present free-text "other" field rather than misleading yes/no buttons.
- **Tests updated** in `test_hitl.py` and `test_orchestrator_helpers_coverage.py` to match the new empty-list fallback behavior.

## Files changed

| File | Change |
|---|---|
| `backend/agents/coding_team/tech_lead_agent/prompts.py` | Options required in schema; rule added with example (auth strategy options) |
| `backend/agents/coding_team/senior_software_engineer_agent/prompts.py` | Schema expanded to show options; CRITICAL rule added with concrete examples |
| `backend/agents/coding_team/hitl.py` | `DEFAULT_CLARIFICATION_OPTIONS = []`; docstring updated |
| `backend/agents/software_engineering_team/orchestrator.py` | `DEFAULT_CLARIFICATION_OPTIONS = []`; docstring updated |
| `backend/agents/coding_team/tests/test_hitl.py` | Assert `options == []` for plain-string questions |
| `backend/agents/software_engineering_team/tests/test_orchestrator_helpers_coverage.py` | Assert `"options" in q` instead of truthy check |

## Test plan

- [x] `pytest agents/coding_team/tests/test_hitl.py agents/coding_team/tests/test_orchestrator_hitl.py` — 56 passed
- [x] `pytest agents/software_engineering_team/tests/test_orchestrator_helpers_coverage.py` — 64 passed
- [ ] Trigger a coding-team job on a GitHub issue lacking spec details; verify clarification questions now show domain-specific options rather than yes/no/not-sure

https://claude.ai/code/session_014ogfp3w9KMzJHRDToRMuep

---
_Generated by [Claude Code](https://claude.ai/code/session_014ogfp3w9KMzJHRDToRMuep)_


backend/agents/software_engineering_team/orchestrator.py
```diff
@@ -227,12 +227,9 @@ def _llm_pause_error(failed: Dict[str, str]) -> str:
...
 
-# Default options for clarification questions
-DEFAULT_CLARIFICATION_OPTIONS = [
-    {"id": "yes", "label": "Yes"},
-    {"id": "no", "label": "No"},
-    {"id": "not_sure", "label": "Not sure / Need more info"},
-]
+# Fallback options for clarification questions. Empty so the UI shows only the always-present
+# free-text "other" field rather than misleading yes/no options for open-ended questions.
+DEFAULT_CLARIFICATION_OPTIONS: List[Dict[str, Any]] = []
```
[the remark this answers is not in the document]
brandonkindred: Fixed in commit 8b56e73. `get_auto_answer_for_job()` now returns `None` early when the question has no options, before constructing the `OpenQuestion` or calling `auto_answer_question()`. The existing API handler already converts `None` to a 500 error, so the auto-answer request fails cleanly rather than returning `selected_option_id="unknown"` which could be accidentally submitted. A new test `test_get_auto_answer_for_job_no_options_returns_none` covers the guard.

---

brandonkindred:

backend/agents/software_engineering_team/orchestrator.py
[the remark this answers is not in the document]
brandonkindred: Fixed in commit 447ef72. A 422 guard was added to both auto-answer endpoints (`/run-team/{job_id}/auto-answer/{question_id}` and `/product-analysis/{job_id}/auto-answer/{question_id}`) immediately after the question is located but before the LLM path is entered. When a question has no selectable options the endpoint now returns HTTP 422 with a message directing the caller to use the `/answers` endpoint with `other_text`, rather than falling through to `get_auto_answer_for_job` and returning a generic 500. Tests added: `test_auto_answer_run_team_422_when_no_options` and `test_auto_answer_product_analysis_422_when_no_options`.

---

backend/agents/coding_team/hitl.py
```diff
@@ -34,13 +34,10 @@
...
-# present an identical answer UI; the UI always offers an "other" free-text option on top.
-DEFAULT_CLARIFICATION_OPTIONS: List[Dict[str, Any]] = [
-    {"id": "yes", "label": "Yes"},
-    {"id": "no", "label": "No"},
-    {"id": "not_sure", "label": "Not sure / Need more info"},
-]
+# Fallback when an agent omits options entirely (should not happen — prompts require context-specific
+# options). Empty list means the UI shows only the always-present "Other (free text)" field, which
+# is more appropriate than forcing yes/no onto open-ended questions like "What API fields are needed?"
+DEFAULT_CLARIFICATION_OPTIONS: List[Dict[str, Any]] = []
```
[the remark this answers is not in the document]
brandonkindred: The empty fallback is intentional for plain-string entry-gate questions. `CodingTeamPlanInput.open_questions` as `List[str]` represents pre-LLM clarifications that arrive before any agent runs — there is no context from which to generate domain-specific options for them at that stage, so yes/no was never semantically correct for these either. Free-text is the right answer mode. The 422 guard added in commit 447ef72 to the auto-answer endpoints ensures that attempting to auto-answer an optionless question surfaces a clear actionable error rather than a confusing 500.

---

backend/agents/coding_team/senior_software_engineer_agent/prompts.py
```diff
@@ -8,12 +8,23 @@
...
+
+CRITICAL — open question options: every open question MUST include at least 2 context-specific answer options in "options". Do NOT use generic yes/no/not-sure options. Options must represent the actual implementation choices for that specific question. Examples: for "How should errors be surfaced to the user?", options might be [{"id": "opt_toast", "label": "Toast notification"}, {"id": "opt_inline", "label": "Inline error message"}, {"id": "opt_modal", "label": "Modal dialog"}]. For "Who owns the copy for this disclaimer?", options might be [{"id": "opt_product", "label": "Product team provides copy"}, {"id": "opt_legal", "label": "Legal team provides copy"}, {"id": "opt_eng", "label": "Engineering drafts copy for review"}]. Mark the most neutral/safe default with "is_default": true."""
```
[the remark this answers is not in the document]
brandonkindred: Fixed in commit 447ef72. Both agent prompts now explicitly prohibit `"other"` as an option ID (it is reserved for the free-text entry field). `_normalize_options` in `hitl.py` also enforces this as a defensive layer: any option arriving with `id="other"` is renamed to `id="other_opt"` with a warning log, preventing the collision regardless of LLM compliance. Test added: `test_normalize_options_renames_reserved_other_id`.

---

brandonkindred:

backend/agents/coding_team/hitl.py
```diff
@@ -68,9 +65,16 @@ def _normalize_options(raw: Any) -> List[Dict[str, Any]]:
     options: List[Dict[str, Any]] = []
     for o in raw or []:
         if isinstance(o, dict) and o.get("id"):
+            opt_id = str(o["id"])
+            if opt_id == "other":
+                # "other" is the reserved synthetic free-text option added by the UI/API;
+                # using it as a structured option id would cause the answer handler to treat
+                # any selection of this option as a free-text response.
+                logger.warning("Option id 'other' is reserved; renaming to 'other_opt'")
+                opt_id = "other_opt"
```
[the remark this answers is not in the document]
brandonkindred: Fixed in commit c30b36f. The rename-to-`other_opt` approach was replaced: `_normalize_options` now **drops** any option with `id="other"` entirely (with a warning log) rather than renaming it to a fixed string. Dropping is safe because the prompt already prohibits `"other"`, so this only fires on non-compliant LLM output — and a dropped option cannot collide with anything. Test updated: `test_normalize_options_drops_reserved_other_id`.

---

backend/agents/coding_team/senior_software_engineer_agent/prompts.py
```diff
@@ -8,12 +8,23 @@
...
+
+CRITICAL — open question options: every open question MUST include at least 2 context-specific answer options in "options". Do NOT use generic yes/no/not-sure options. Options must represent the actual implementation choices for that specific question. Examples: for "How should errors be surfaced to the user?", options might be [{"id": "opt_toast", "label": "Toast notification"}, {"id": "opt_inline", "label": "Inline error message"}, {"id": "opt_modal", "label": "Modal dialog"}]. For "Who owns the copy for this disclaimer?", options might be [{"id": "opt_product", "label": "Product team provides copy"}, {"id": "opt_legal", "label": "Legal team provides copy"}, {"id": "opt_eng", "label": "Engineering drafts copy for review"}]. Mark the most neutral/safe default with "is_default": true. NEVER use "other" as an option id — it is reserved for the free-text entry field."""
```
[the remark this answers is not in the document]
brandonkindred: Fixed in commit c30b36f. `normalize_open_questions` now discards options when fewer than 2 survive normalization (falling back to free-text with a warning log), enforcing the ≥2 requirement deterministically regardless of LLM compliance. The commit also addressed the P1 raised in the review body (orchestrator.py line 273): the `/run-team/{job_id}/answers` endpoint now validates that submitted option IDs are actually present in the question's options list, and that optionless questions receive non-blank `other_text` — preventing empty or fabricated answers from clearing the pause gate. Tests added: `test_normalize_open_questions_drops_options_fewer_than_two`, `test_submit_pending_answers_400_when_unknown_option_id`, `test_submit_pending_answers_400_when_no_option_and_no_text`.

---

backend/agents/coding_team/hitl.py
```diff
@@ -148,8 +153,17 @@ def normalize_open_questions(raw: Any) -> List[Dict[str, Any]]:
             if q.get("context"):
                 entry["context"] = str(q["context"])
             opts = _normalize_options(q.get("options"))
-            if opts:
+            # Fewer than 2 options after normalization indicates a non-compliant LLM
+            # response (e.g. single option, or "other" dropped). Fall back to free-text.
+            if len(opts) >= 2:
```
brandonkindred: **[CRITICAL] logic** — In convert_to_structured_questions, when processing dict questions, the code only sets the 'options' key if len(opts) >= 2 (line 158-159). If opts is empty or has 1 item, no 'options' key is added to the entry dict. This violates the function's own docstring which states 'else an empty list', and causes the test test_normalize_options_drops_reserved_other_id (which expects out[0]['options'] == []) to fail with a KeyError. The same issue likely affects plain-string questions (the string branch is not shown but presumably also omits the key). The UI and downstream code expect the 'options' field to be present (even if empty) to distinguish 'no options' from 'missing key'. This will break the HITL flow when questions have no valid options.

**Suggested fix:** Modify the function to always include an 'options' key. For plain-string questions, initialize entry with 'options': []. For dict questions, after normalization, set entry['options'] = opts if len(opts) >= 2, else set entry['options'] = [] (and log a warning if opts had 1 item). Example fix for the dict branch (around line 158):

opts = _normalize_options(q.get('options'))
if len(opts) >= 2:
    entry['options'] = opts
else:
    if opts:
        logger.warning(...)
    entry['options'] = []

Ensure the string branch (not shown) also sets 'options': [].

brandonkindred (commented):
The changes correctly update prompts to require context-specific options and set fallback to empty list. However, a critical bug in convert_to_structured_questions causes the 'options' key to be omitted when no valid options exist, breaking tests and potentially the UI. This must be fixed before merge. Additionally, there is an inconsistency between two normalization functions that should be resolved for maintainability.

**Spec compliance:** The prompts now require options, and the fallback is empty as specified. Tests are updated to reflect the new behavior, but the implementation bug in hitl.py prevents the tests from passing. Once fixed, the code will meet the acceptance criteria.

- `backend/agents/coding_team/hitl.py` — **[MEDIUM] standards** — The function normalize_open_questions (tested in test_hitl.py) omits the 'options' key when options are insufficient (test expects 'options' not in out[0]), while convert_to_structured_questions is intended to always include an empty list. This inconsistency in output format between two normalization functions in the same module may confuse maintainers and lead to subtle bugs if the functions are used interchangeably. The API handles missing keys gracefully, but a uniform contract (always include 'options' key, even if empty) would be clearer.

**Suggested fix:** Decide on a single contract for question dictionaries: either always include 'options' (empty list when none) or always omit the key when empty. Update both functions and their tests to follow the same convention, and document the contract in the module docstring.

backend/agents/coding_team/hitl.py
```diff
@@ -68,9 +65,17 @@ def _normalize_options(raw: Any) -> List[Dict[str, Any]]:
     options: List[Dict[str, Any]] = []
     for o in raw or []:
         if isinstance(o, dict) and o.get("id"):
+            opt_id = str(o["id"])
```
brandonkindred: **[INFO] standards** — In _normalize_options, the code accesses o['id'] directly (line 68: opt_id = str(o['id'])). If an option dict is missing the 'id' key (e.g., due to a malformed LLM response), this will raise a KeyError and crash the normalization. The prompt requires 'id', but defensive coding would guard against missing keys.

**Suggested fix:** Use o.get('id') and skip the option with a warning if 'id' is missing: opt_id = o.get('id'); if not opt_id: logger.warning('Option missing id, skipping'); continue

backend/agents/coding_team/hitl.py
brandonkindred: Fixed in commit 458e697. `normalize_open_questions` now always sets `entry["options"]` — plain-string questions get `options: []` via a new `else` branch, and dict questions with fewer than 2 normalized options also fall through to `entry["options"] = []` (with a warning log when 1 option was discarded). The docstring postcondition was updated to reflect the uniform contract: every returned dict carries an `options` key. Tests updated: `test_normalize_open_questions_drops_options_fewer_than_two` now asserts `out[0]["options"] == []`, and `test_normalize_open_questions_strings_and_dicts` now asserts `out[0] == {"question_text": "plain?", "options": []}`.

---

backend/agents/coding_team/hitl.py
brandonkindred: Fixed in commit 458e697. `_normalize_options` now uses `o.get("id")` (line 68: `opt_id = str(o.get("id"))`) instead of the direct `o["id"]` access. The existing guard `if isinstance(o, dict) and o.get("id"):` already prevents a `None` from reaching `str()`, so this is style-only hardening — but it removes any ambiguity about intent.

---

brandonkindred:
@codex review

backend/agents/software_engineering_team/api/main.py
```diff
@@ -2157,6 +2173,12 @@ def auto_answer_run_team_question(
             detail=f"Question {question_id} not found in pending questions.",
         )
 
+    if not question_data.get("options"):
+        raise HTTPException(
+            status_code=422,
+            detail="This question has no selectable options. Provide a free-text answer via the /answers endpoint using the other_text field.",
```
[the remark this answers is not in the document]
brandonkindred: Fixed in commit a196fea. The auto-answer button in `question-card.component.html` is now wrapped in `@if (question.options?.length)` so it is only rendered when the question has at least one selectable option. Optionless questions show only the always-present free-text "Other (specify below)" control, which is the correct answer path for them.

---

brandonkindred:

backend/agents/software_engineering_team/product_requirements_analysis_agent/auto_answer.py
```diff
@@ -218,6 +218,14 @@ def get_auto_answer_for_job(
...
     ]
 
+    if not options:
+        # Auto-answer requires at least one option to select from; questions without options
+        # must be answered by the user via free-text (the "other" field).
+        logger.warning(
+            "Question %s in job %s has no options; cannot auto-answer", question_id, job_id
+        )
+        return None
+
```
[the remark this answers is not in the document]
brandonkindred: Fixed in commit 8453c24. `get_auto_answer_for_job` now filters out options with `id == "other"` before the `if not options:` guard, so the synthetic `{"id": "other", "label": "Provide answer in text field"}` placeholder inserted by `_convert_to_pending_questions` is invisible to the LLM path. A question whose only option is that placeholder is now treated as fully optionless and returns `None`. Test added: `test_get_auto_answer_for_job_synthetic_other_option_returns_none`.

---

backend/agents/coding_team/hitl.py
```diff
@@ -148,8 +155,19 @@ def normalize_open_questions(raw: Any) -> List[Dict[str, Any]]:
             if q.get("context"):
                 entry["context"] = str(q["context"])
             opts = _normalize_options(q.get("options"))
-            if opts:
+            if len(opts) >= 2:
                 entry["options"] = opts
```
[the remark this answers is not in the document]
brandonkindred: Fixed in commit 8453c24. The check was changed from `len(opts) >= 2` to `len({o["id"] for o in opts}) >= 2`, counting distinct IDs instead of raw list length. Two options sharing the same ID now correctly fall back to free-text with a warning log rather than being surfaced as two selectable choices. Test added: `test_normalize_open_questions_drops_duplicate_option_ids`.

---

backend/agents/coding_team/hitl.py
```diff
@@ -148,8 +155,19 @@ def normalize_open_questions(raw: Any) -> List[Dict[str, Any]]:
...
+            else:
```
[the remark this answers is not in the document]
brandonkindred: Fixed in commit 8453c24. A `_GENERIC_OPTION_IDS = frozenset({"yes", "no", "not_sure"})` constant was added to `hitl.py`. The acceptance condition in `normalize_open_questions` is now `len(unique_ids) >= 2 and not unique_ids.issubset(_GENERIC_OPTION_IDS)` — so a non-compliant LLM that emits the former yes/no/not-sure defaults is rejected deterministically and the question falls back to free-text with a distinct warning log message. Test added: `test_normalize_open_questions_drops_generic_yes_no_options`.

---

brandonkindred:

backend/agents/coding_team/hitl.py
```diff
@@ -148,8 +160,27 @@ def normalize_open_questions(raw: Any) -> List[Dict[str, Any]]:
             if q.get("context"):
                 entry["context"] = str(q["context"])
             opts = _normalize_options(q.get("options"))
-            if opts:
+            unique_ids = {o["id"] for o in opts}
+            if len(unique_ids) >= 2 and not unique_ids.issubset(_GENERIC_OPTION_IDS):
```
brandonkindred: **[MEDIUM] logic** — The check `unique_ids.issubset(_GENERIC_OPTION_IDS)` discards options if all option IDs are in {"yes", "no", "not_sure"}. While the prompt instructs the LLM to avoid these IDs, a non-compliant LLM could still emit options with these IDs but with context-specific labels (e.g., {"id": "yes", "label": "Yes, use OAuth2"}, {"id": "no", "label": "No, use local auth"}). The current logic would discard these options and fall back to free-text, potentially degrading the user experience.

**Suggested fix:** Consider also checking whether the labels are the generic "Yes"/"No"/"Not sure" strings before discarding, or log a warning and keep the options if labels appear context-specific. Alternatively, accept that the prompt enforcement is sufficient and document the rationale.

user-interface/src/app/components/pending-questions/question-card/question-card.component.html
```diff
@@ -5,19 +5,21 @@
...
-      (click)="onAutoAnswerRequest()"
-      matTooltip="Auto-answer using AI based on industry best practices"
-    >
-      @if (isAutoAnswering) {
-        <mat-spinner diameter="20"></mat-spinner>
-      } @else {
-        <mat-icon>auto_awesome</mat-icon>
-      }
-    </button>
+    @if (question.options?.length) {
```
brandonkindred: **[INFO] integration** — The auto-answer button is hidden when `question.options?.length` is falsy, which correctly prevents auto-answer on optionless questions. The free-text "other" field is always rendered (not shown in this snippet), ensuring users can still provide free-text answers. This is consistent with the backend behavior.

**Suggested fix:** No action required.

brandonkindred (commented):
The PR successfully implements the requirement to enforce context-specific answer options for coding team clarification questions. The prompts are updated, the fallback is changed to an empty list, and the UI hides the auto-answer button when no options are available. The code is well-structured, with proper error handling, logging, and documentation. Tests cover the main scenarios. Minor suggestions: consider refining the generic ID filter to avoid discarding context-specific labels, and add a test for missing options key.

**Spec compliance:** The code meets the acceptance criteria: prompts require ≥2 context-specific options, DEFAULT_CLARIFICATION_OPTIONS is empty, tests updated, and UI adjusted. The manual test step (trigger a coding-team job) is not automated but is outside the scope of code review.

- `backend/agents/coding_team/tests/test_hitl.py` — **[LOW] testing** — There is no test case for `normalize_open_questions` when the input dict is missing the `options` key entirely. The function handles it gracefully (returns empty options), but a test would ensure this behavior doesn't regress.

**Suggested fix:** Add a test like `def test_normalize_open_questions_missing_options_key(): out = hitl.normalize_open_questions([{"question_text": "Q?"}]); assert out[0]["options"] == []`.

backend/agents/coding_team/hitl.py
brandonkindred: Acknowledged and documented. The `_GENERIC_OPTION_IDS` check is intentionally narrow: it only discards options when *all* IDs are in the known-bad set `{"yes", "no", "not_sure"}` — the exact IDs the old fallback constant used. Options with those IDs but context-specific labels (e.g. `{"id": "yes", "label": "Yes, use OAuth2"}`) are an edge case that shouldn't occur in practice because the prompts explicitly require non-generic IDs. A label-based detector would add significant complexity (what patterns count as generic?) and could incorrectly reject legitimate options. The primary enforcement mechanism is prompt compliance; the ID check is a narrow backstop against the specific known failure mode. A comment explaining this tradeoff has been added to the `_GENERIC_OPTION_IDS` constant in commit e899d42.

---

user-interface/src/app/components/pending-questions/question-card/question-card.component.html
brandonkindred: Confirmed — no action taken. The `@if (question.options?.some(o => o.id !== 'other'))` guard (commit e899d42) supersedes the previous `?.length` check and correctly handles both empty lists and synthetic-only lists.

---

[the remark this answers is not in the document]
brandonkindred: Fixed in commit e899d42. Both auto-answer API guards (`/run-team` and `/product-analysis`) now build a `real_options` list that excludes options with `id == "other"` before testing emptiness. The UI guard was also tightened from `question.options?.length` to `question.options?.some(o => o.id !== 'other')` so the auto-answer button is hidden whenever the only option is the synthetic placeholder. Test added: `test_auto_answer_run_team_422_when_only_synthetic_other_option`.

---

backend/agents/coding_team/hitl.py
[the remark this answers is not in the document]
brandonkindred: Declined — label inspection is intentionally not implemented here. The `_GENERIC_OPTION_IDS` guard targets the specific known-bad pattern (the old fallback constant's exact IDs), not genericness in general. Label-based detection would require heuristics that could incorrectly reject legitimate options (e.g. a deployment option labelled "No downtime required"). The prompt prohibition on generic IDs is the primary enforcement; any LLM that ignores it and uses generic IDs gets caught by the narrow ID set, while one that uses generic labels with different IDs is an extremely unlikely non-compliance scenario that falls outside this defensive guard's scope.

---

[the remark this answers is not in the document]
brandonkindred: Fixed in commit e899d42. `normalize_open_questions` now deduplicates `opts` by ID (first occurrence wins) before the unique-ID count check. `[a, a, b]` deduplicates to `[a, b]` and is accepted cleanly with no duplicate in the rendered set; `[a, a]` deduplicates to `[a]` — one unique ID — and falls back to free-text as before. Tests added: `test_normalize_open_questions_deduplicates_partial_duplicates` verifies the `[a, a, b]` → `[a, b]` path.

---

brandonkindred:

backend/agents/coding_team/hitl.py
```diff
@@ -68,9 +70,17 @@ def _normalize_options(raw: Any) -> List[Dict[str, Any]]:
     options: List[Dict[str, Any]] = []
     for o in raw or []:
         if isinstance(o, dict) and o.get("id"):
+            opt_id = str(o.get("id"))
+            if opt_id == "other":
```
brandonkindred: **[MEDIUM] logic** — The reserved option id check `if opt_id == "other"` is case-sensitive. If a non-compliant LLM emits an option with id "Other" or "OTHER", it will not be dropped, potentially allowing the reserved free-text placeholder to be treated as a real selectable option.

**Suggested fix:** Use case-insensitive comparison: `if opt_id.lower() == "other":`

backend/agents/coding_team/hitl.py
```diff
@@ -148,8 +160,38 @@ def normalize_open_questions(raw: Any) -> List[Dict[str, Any]]:
...
+            # both rendering and recorded decisions.
+            seen_ids: set = set()
+            deduped: list = []
+            for o in opts:
+                if o["id"] not in seen_ids:
+                    seen_ids.add(o["id"])
+                    deduped.append(o)
+            opts = deduped
+            unique_ids = {o["id"] for o in opts}
+            if len(unique_ids) >= 2 and not unique_ids.issubset(_GENERIC_OPTION_IDS):
```
brandonkindred: **[MEDIUM] logic** — The generic option IDs check `unique_ids.issubset(_GENERIC_OPTION_IDS)` uses a lowercase frozenset. If the LLM emits "Yes", "No", etc. (capitalized), they will not be detected as generic and may surface as context-irrelevant choices.

**Suggested fix:** Normalize IDs to lowercase before checking: `unique_ids_lower = {o.lower() for o in unique_ids}` and compare against a lowercase frozenset, or make `_GENERIC_OPTION_IDS` case-insensitive by storing lowered values and lowering IDs at comparison time.

backend/agents/coding_team/hitl.py
```diff
@@ -132,9 +142,11 @@ def normalize_open_questions(raw: Any) -> List[Dict[str, Any]]:
     Preconditions:
         - ``raw`` is a list of strings or dicts (or None).
     Postconditions:
-        - Returns dicts each carrying at least ``question_text``; empties are dropped, and any
-          ``context`` / ``options`` an agent supplied are preserved so domain-specific choices
-          round-trip. A non-list input yields ``[]``.
+        - Returns dicts each carrying ``question_text`` and ``options`` (always present); empties
+          are dropped, and any ``context`` / ``options`` an agent supplied are preserved so
+          domain-specific choices round-trip. Options that fail normalization (< 2 survive) are
+          discarded and ``options`` is set to ``[]`` so callers have a uniform contract — the same
```
brandonkindred: **[MEDIUM] documentation** — The docstring for `normalize_open_questions` states options are discarded when fewer than 2 survive, but does not mention that options consisting solely of generic yes/no/not-sure IDs are also discarded.

**Suggested fix:** Update the postcondition description to include: "or consist exclusively of generic yes/no/not-sure IDs".

backend/agents/software_engineering_team/api/main.py
```diff
@@ -2157,6 +2173,15 @@ def auto_answer_run_team_question(
             detail=f"Question {question_id} not found in pending questions.",
         )
 
+    # Filter out the synthetic {"id": "other"} placeholder before checking for real options;
+    # _convert_to_pending_questions inserts it when a question has no structured options.
+    real_options = [o for o in (question_data.get("options") or []) if o.get("id") != "other"]
```
brandonkindred: **[MEDIUM] logic** — The filter `if o.get("id") != "other"` is case-sensitive. A non-compliant LLM could emit an option with id "Other" that bypasses this guard and is treated as a real option, potentially breaking the auto-answer endpoint's optionless detection.

**Suggested fix:** Use case-insensitive comparison: `if (o.get("id") or "").lower() != "other"`

backend/agents/software_engineering_team/api/main.py
```diff
@@ -2659,6 +2684,15 @@ def auto_answer_product_analysis_question(
...
```
brandonkindred: **[MEDIUM] logic** — Same case-sensitive "other" check as line 2178 (duplicate endpoint logic).

**Suggested fix:** Apply the same case-insensitive fix: `if (o.get("id") or "").lower() != "other"`

backend/agents/software_engineering_team/product_requirements_analysis_agent/auto_answer.py
```diff
@@ -216,8 +216,22 @@ def get_auto_answer_for_job(
             confidence=float(opt.get("confidence", 0.0)),
         )
         for i, opt in enumerate(question_data.get("options", []))
+        # Filter out the synthetic "other" placeholder: _convert_to_pending_questions inserts
+        # {"id": "other", ...} when a question has no real options. Including it here would let
+        # the LLM "select" the placeholder and record it as the user's decision without any
+        # actual user input. Treat questions whose only option is the synthetic placeholder the
+        # same as fully optionless questions.
+        if opt.get("id") != "other"
```
brandonkindred: **[MEDIUM] logic** — The filter `if opt.get("id") != "other"` is case-sensitive, creating the same risk of a non-lowercase "other" variant being treated as a real option and allowing the LLM to select the placeholder.

**Suggested fix:** Use case-insensitive comparison: `if (opt.get("id") or "").lower() != "other"`

user-interface/src/app/components/pending-questions/question-card/question-card.component.ts
```diff
@@ -50,6 +50,11 @@ export class QuestionCardComponent {
     return this.question.allow_multiple === true;
   }
 
+  /** True when the question has at least one real selectable option (not the synthetic "other" placeholder). */
+  get hasSelectableOptions(): boolean {
+    return (this.question?.options ?? []).some(o => o.id !== 'other');
```
brandonkindred: **[LOW] logic** — The `hasSelectableOptions` getter uses `o.id !== 'other'` (case-sensitive). While the backend currently inserts lowercase "other", a case-insensitive check would be more robust against future changes or unexpected data.

**Suggested fix:** Use `o.id.toLowerCase() !== 'other'` for defensive consistency with backend guards.

brandonkindred (commented):
The changes correctly implement the requirement to enforce context-specific answer options and fall back to free-text when options are missing or generic. Code is well-structured, tests are updated and passing. Minor robustness improvements are suggested: making reserved/generic ID checks case-insensitive to harden against non-compliant LLM outputs, and updating a docstring to fully describe the discard conditions.

**Spec compliance:** The implementation meets the acceptance criteria: prompts now require ≥2 context-specific options with examples; DEFAULT_CLARIFICATION_OPTIONS is empty so the UI falls back to free-text; tests verify the new behavior. The UI correctly hides the auto-answer button when no real options exist. The manual test scenario (trigger a coding-team job on an underspecified issue) should now produce domain-specific options instead of yes/no/not-sure.

backend/agents/coding_team/hitl.py
brandonkindred: Fixed in commit 1204c90. `_normalize_options` now uses `opt_id.lower() == "other"` so mixed-case variants emitted by a non-compliant LLM (`"Other"`, `"OTHER"`) are dropped by the same guard.

---

backend/agents/coding_team/hitl.py
brandonkindred: Fixed in commit 1204c90. The whole-set subset check has been replaced with per-option filtering (see the new P2 resolution below), which makes the lowercase normalization moot — each option's ID and label are now tested individually with `.lower()` before the minimum-count check, so capitalized IDs (`"Yes"`, `"NO"`) are caught regardless.

---

backend/agents/coding_team/hitl.py
brandonkindred: Fixed in commit 1204c90. The `normalize_open_questions` docstring postcondition now reads: "Options that fail normalization (< 2 survive after deduplication, or consist exclusively of generic yes/no/not-sure IDs) are discarded and `options` is set to `[]`."

---

backend/agents/software_engineering_team/api/main.py
brandonkindred: Fixed in commit 1204c90. Both auto-answer endpoints now use `(o.get("id") or "").lower() != "other"` so the synthetic placeholder is filtered out regardless of case.

---

backend/agents/software_engineering_team/api/main.py
brandonkindred: Fixed in commit 1204c90 (same as line 2178 above — `replace_all=true` updated both occurrences together).

---

backend/agents/software_engineering_team/product_requirements_analysis_agent/auto_answer.py
brandonkindred: Fixed in commit 1204c90. `auto_answer.py` now uses `(opt.get("id") or "").lower() != "other"`.

---

user-interface/src/app/components/pending-questions/question-card/question-card.component.ts
brandonkindred: Fixed in commit 1204c90. `hasSelectableOptions` now uses `o.id.toLowerCase() !== 'other'`.

---

backend/agents/coding_team/hitl.py
```diff
@@ -148,8 +160,38 @@ def normalize_open_questions(raw: Any) -> List[Dict[str, Any]]:
...
+            seen_ids: set = set()
+            deduped: list = []
+            for o in opts:
+                if o["id"] not in seen_ids:
+                    seen_ids.add(o["id"])
+                    deduped.append(o)
+            opts = deduped
+            unique_ids = {o["id"] for o in opts}
+            if len(unique_ids) >= 2 and not unique_ids.issubset(_GENERIC_OPTION_IDS):
                 entry["options"] = opts
```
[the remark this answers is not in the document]
brandonkindred: Fixed in commit 1204c90. The whole-set subset check was replaced with per-option filtering:

- Added `_GENERIC_OPTION_LABELS = frozenset({"yes", "no", "not sure", "not_sure", "not sure / need more info"})`.
- Each option is now tested individually: an option is culled if its `id.lower()` is in `_GENERIC_OPTION_IDS` **or** its `label.lower()` is in `_GENERIC_OPTION_LABELS`.
- After culling, if ≥ 2 options remain they are accepted; otherwise the question falls back to free-text.

This handles both failure modes: mixed sets (`yes`/`no` + one context-specific option → generic ones removed, specific kept) and variant IDs recognized by label (`id="opt_yes", label="Yes"` → culled by label match). Tests added: `test_normalize_open_questions_drops_generic_mixed_with_specific`, `test_normalize_open_questions_drops_variant_ids_by_label`, and `test_normalize_open_questions_mixed_keeps_specific_when_two_or_more`.

---

brandonkindred:

backend/agents/coding_team/hitl.py
```diff
@@ -68,9 +72,17 @@ def _normalize_options(raw: Any) -> List[Dict[str, Any]]:
     options: List[Dict[str, Any]] = []
     for o in raw or []:
         if isinstance(o, dict) and o.get("id"):
+            opt_id = str(o.get("id"))
+            if opt_id.lower() == "other":
```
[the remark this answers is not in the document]
brandonkindred: Fixed in commit f83bf14. `_normalize_options` now calls `.strip()` on the raw ID immediately after `str()` conversion, skips whitespace-only IDs (`if not opt_id: continue`), and strips the label at storage time. This means `"other "` → stripped to `"other"` → dropped by the reserved-ID guard, and `" Yes "` → stripped to `"Yes"` → matched against `_GENERIC_OPTION_LABELS` and culled. Test added: `test_normalize_options_strips_whitespace_from_id_and_label`.

---

brandonkindred:

brandonkindred:

backend/agents/coding_team/hitl.py
```diff
@@ -34,13 +34,17 @@
...
+# options). Empty list means the UI shows only the always-present "Other (free text)" field, which
+# is more appropriate than forcing yes/no onto open-ended questions like "What API fields are needed?"
+DEFAULT_CLARIFICATION_OPTIONS: List[Dict[str, Any]] = []
+
+# Option IDs and labels that signal a generic yes/no/not-sure response pattern. Individual options
+# matching either set are culled before the minimum-count check so that mixed sets (e.g. "yes"/"no"
+# blended with one context-specific option) and variant IDs recognized by their display label
+# (e.g. id="opt_yes", label="Yes") are both removed rather than silently accepted.
+_GENERIC_OPTION_IDS: frozenset = frozenset({"yes", "no", "not_sure"})
+_GENERIC_OPTION_LABELS: frozenset = frozenset({"yes", "no", "not sure", "not_sure", "not sure / need more info"})
```
brandonkindred: **[CONFIRMED] `_GENERIC_OPTION_LABELS` is missing "other" variants — a non-compliant LLM can smuggle a fake free-text option into the structured list**

`_normalize_options` only blocks options where `id.lower() == "other"`. The label filter in `normalize_open_questions` only rejects labels in `_GENERIC_OPTION_LABELS`. An LLM that emits `{"id": "choice_other", "label": "Other"}` (non-reserved id, generic label) passes both filters and reaches the UI as a fake structured radio/checkbox option.

If two such options exist they satisfy the `>= 2` threshold and appear as selectable choices. When selected, `answers_to_resolved` records `"choice_other"` as the structured answer instead of routing to the free-text path.

```suggestion
_GENERIC_OPTION_LABELS: frozenset = frozenset({
    "yes", "no", "not sure", "not_sure", "not sure / need more info",
    "other", "other (specify)", "other (please specify)", "other (free text)",
})
```

---

backend/agents/coding_team/hitl.py
```diff
@@ -148,8 +165,45 @@ def normalize_open_questions(raw: Any) -> List[Dict[str, Any]]:
...
+            # `track option.id` and logs NG0955 for duplicates; answer resolution also
+            # picks the first label matching a submitted ID, so duplicate IDs corrupt
+            # both rendering and recorded decisions.
+            seen_ids: set = set()
+            deduped: list = []
+            for o in opts:
+                if o["id"] not in seen_ids:
+                    seen_ids.add(o["id"])
+                    deduped.append(o)
+            opts = deduped
```
brandonkindred: **[PLAUSIBLE] Dedup uses case-sensitive key but `answers_to_resolved` compares IDs case-insensitively — two IDs like `"Cloud"` / `"cloud"` both survive dedup, then the wrong label is recorded**

The dedup loop at this line tracks `o["id"]` with exact-string equality, so an LLM that emits both `"Cloud"` and `"cloud"` as option IDs preserves both in `opts`. `answers_to_resolved` (line 278) was changed in this PR to use `(opt.get("id") or "").lower() == selected.lower()`. If the user selects the second option (`"cloud"`), the loop finds the first option (`"Cloud"`) first and returns its label, silently recording the wrong decision.

Fix: use lowercased key in the dedup loop so case-variant IDs are deduplicated:

```suggestion
            seen_ids: set = set()
            deduped: list = []
            for o in opts:
                if o["id"].lower() not in seen_ids:
                    seen_ids.add(o["id"].lower())
                    deduped.append(o)
            opts = deduped
```

---

backend/agents/coding_team/hitl.py
```diff
@@ -68,10 +72,20 @@ def _normalize_options(raw: Any) -> List[Dict[str, Any]]:
     options: List[Dict[str, Any]] = []
     for o in raw or []:
         if isinstance(o, dict) and o.get("id"):
+            opt_id = str(o.get("id")).strip()
```
brandonkindred, line 84: **[PLAUSIBLE] `convert_to_structured_questions` trims IDs but never deduplicates — two inputs `"a"` and `" a "` both normalize to `"a"`, producing duplicate Angular tracking keys**

`_normalize_options` is called by both `convert_to_structured_questions` (dict-question path, line 124) and `normalize_open_questions` (line 165). The deduplication loop that prevents Angular NG0955 was added only inside `normalize_open_questions`. If the same question passes through `convert_to_structured_questions` with options `[{"id":"a"}, {"id":" a "}]`, both normalize to `id="a"` and both survive — the caller receives a list with two identical IDs.

Moving the dedup inside `_normalize_options` fixes both paths at once and removes the need for the separate loop in `normalize_open_questions`:

```python
def _normalize_options(raw: Any) -> List[Dict[str, Any]]:
    options: List[Dict[str, Any]] = []
    seen_ids: set = set()
    for o in raw or []:
        if isinstance(o, dict) and o.get("id"):
            opt_id = str(o.get("id")).strip()
            if not opt_id:
                logger.warning("Option with whitespace-only id will be dropped")
                continue
            if opt_id.lower() == "other":
                logger.warning("Option id 'other' is reserved and will be dropped")
                continue
            norm_key = opt_id.lower()
            if norm_key in seen_ids:
                logger.warning("Duplicate option id '%s' will be dropped", opt_id)
                continue
            seen_ids.add(norm_key)
            options.append({"id": opt_id, "label": str(o.get("label") or opt_id).strip(), "is_default": bool(o.get("is_default", False))})
    return options
```

---

backend/agents/coding_team/hitl.py
```diff
@@ -68,10 +72,20 @@ def _normalize_options(raw: Any) -> List[Dict[str, Any]]:
...
+            if not opt_id:
```
brandonkindred, line 85: **[CONFIRMED] Whitespace-only option ID is silently dropped with no log warning**

A whitespace-only ID like `"   "` passes the outer `o.get("id")` truthiness check (non-empty string), then `str(...).strip()` collapses it to `""`, triggering `continue` here — but unlike the `"other"` case on line 83, there is no `logger.warning` call. A non-compliant LLM emitting `{"id": "   ", "label": "Foo"}` silently discards the option with no trace in logs.

```suggestion
            if not opt_id:
                logger.warning("Option with whitespace-only id will be dropped")
                continue
```

---

brandonkindred (commented):
Code review (high effort, recall-biased) — 4 surviving findings after verification. Two CONFIRMED bugs, two PLAUSIBLE.\n\nAlso addressing the two Codex inline comments below (both map to findings #2 and #3 in this review). Will fix all four in a follow-up commit."

---

brandonkindred:

backend/agents/coding_team/hitl.py
```diff
@@ -107,7 +136,7 @@ def convert_to_structured_questions(
             continue
         if isinstance(q, dict):
             qid = str(q.get("id") or f"{source}_{idx}_{uuid.uuid4().hex[:8]}")
-            options = _normalize_options(q.get("options")) or list(DEFAULT_CLARIFICATION_OPTIONS)
+            options = _normalize_options(q.get("options"))
```
brandonkindred: **[MEDIUM] logic** — The `convert_to_structured_questions` function does not apply the generic option filtering (using `_GENERIC_OPTION_IDS` and `_GENERIC_OPTION_LABELS`) that `normalize_open_questions` applies. This means if an agent using this conversion path (e.g., the software engineering team's orchestrator for planning questions) emits generic options like yes/no/not-sure, they will be accepted instead of being discarded. This undermines the defense-in-depth goal of eliminating generic options across all teams.

**Suggested fix:** Add filtering logic similar to `normalize_open_questions` after the `_normalize_options` call. For example, after line 139, filter `options` to remove entries whose id or label is generic, and if fewer than 2 remain, set `options = []`. Alternatively, refactor the filtering into a shared helper used by both functions.

backend/agents/coding_team/hitl.py
```diff
@@ -95,10 +124,10 @@ def convert_to_structured_questions(
           question text (``question_text`` / ``text`` / ``question``).
```
brandonkindred, line 153: **[LOW] documentation** — The docstring for `convert_to_structured_questions` does not mention the expected behavior regarding generic options, and it is inconsistent with `normalize_open_questions` which explicitly documents filtering. This may confuse maintainers about whether generic options are handled.

**Suggested fix:** Update the docstring to clarify that generic options are not filtered in this function (or, if filtering is added, document the filtering behavior). Align the documentation with the actual implementation.

backend/agents/coding_team/tech_lead_agent/prompts.py
```diff
@@ -25,7 +26,7 @@
 Respond with a single JSON object with keys "tasks", "stacks", and "open_questions".
 "tasks": list of {{ "id": str, "title": str, "description": str, "dependencies": list[str] }}
 "stacks": list of {{ "name": str, "tools_services": list[str] }}
-"open_questions": list of {{ "question_text": str, "context": str (optional), "options": list[{{ "id": str, "label": str }}] (optional) }}"""
+"open_questions": list of {{ "question_text": str, "context": str, "options": list[{{ "id": str, "label": str, "is_default": bool }}] }}"""
```
brandonkindred, line 29: **[LOW] logic** — The prompt string uses double curly braces `{{` and `}}` for the JSON schema example. If this string is used as a system prompt without `.format()`, the LLM will see literal `{{` instead of `{`, which could cause confusion or malformed JSON understanding. Verify whether this string is intended to be formatted; if not, the escaping is incorrect.

**Suggested fix:** Check how the prompt variable is used. If it is passed directly to the LLM, replace `{{` with `{` and `}}` with `}`. If it is used with `.format()`, ensure that the formatting is correct and that no unintended substitutions occur.

brandonkindred (commented):
The PR correctly implements the requirement for context-specific answer options in the coding team's clarification questions. Prompts are updated, fallback defaults are set to empty lists, and tests are comprehensive. The main area for improvement is the inconsistency in generic option filtering between the two question normalization functions, which could allow generic options to slip through in the planning team path. Otherwise, the code is well-structured, documented, and tested.

**Spec compliance:** The code meets the acceptance criteria: coding team prompts now require ≥2 context-specific options with examples, DEFAULT_CLARIFICATION_OPTIONS is empty, and tests verify the new behavior. The planning team path (convert_to_structured_questions) does not filter generic options, which is a minor gap but does not affect the coding team flow.

- `backend/agents/coding_team/tests/test_hitl.py` — **[LOW] testing** — There is no test that verifies the behavior of `convert_to_structured_questions` when provided with generic options (e.g., yes/no). Since the function does not filter them, a test would document the expected behavior and catch regressions if filtering is later added or if the behavior changes unintentionally.

**Suggested fix:** Add a test case such as `test_convert_to_structured_questions_keeps_generic_options` that passes generic options and asserts they are preserved (or, if filtering is added, that they are removed).

backend/agents/coding_team/tech_lead_agent/prompts.py
brandonkindred: The double braces are correct here. `PLAN_TO_TASK_GRAPH_USER` is used with `.format(plan_text=plan_text)` in `tech_lead_agent/agent.py` (line 133), so `{{` and `}}` are standard Python escape sequences for literal `{` and `}` in a format string — the LLM receives the correct single-brace JSON. No change needed.

---

backend/agents/coding_team/hitl.py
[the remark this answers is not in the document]
brandonkindred: Fixed in commit `7b2c324`. Deduplication was moved inside `_normalize_options` using a **case-insensitive** key (`norm_key = opt_id.lower()`), so it now applies to every caller including `convert_to_structured_questions`. IDs `"a"` and `" a "` both strip to `"a"` and the second is dropped with a warning log. A test `test_normalize_options_deduplicates_within_convert_to_structured_questions` covers this path.

---

brandonkindred: Fixed in commit `7b2c324`. Deduplication was moved from the inline loop in `normalize_open_questions` into `_normalize_options` itself, using a lowercased key — so both `convert_to_structured_questions` and `normalize_open_questions` now deduplicate through the same path. `test_normalize_options_deduplicates_within_convert_to_structured_questions` covers this.

---

backend/agents/coding_team/hitl.py
brandonkindred: Fixed in commit `7b2c324`. Added `logger.warning("Option with whitespace-only id will be dropped")` before the `continue`, consistent with the existing warning for reserved `"other"` IDs.

---

backend/agents/coding_team/hitl.py
brandonkindred: Fixed in commit `eab5ef3`. The docstring for `convert_to_structured_questions` was updated to document that `_filter_generic_options` is applied and that no minimum count is enforced (unlike `normalize_open_questions`). A shared `_filter_generic_options` helper was also extracted so both functions use the same filtering logic.

---

backend/agents/coding_team/tech_lead_agent/prompts.py
```diff
@@ -16,6 +16,7 @@
...
 - Open questions: if any required product/design decision is missing, put it in "open_questions" and you may return an empty "tasks" list — the job pauses for a human to answer, then you will be re-asked with the answers.
+- Open question options: every open question MUST include at least 2 context-specific answer options. Do NOT use generic yes/no/not-sure options. Options must represent the actual choices available for that specific decision. For example, if asking "Which auth strategy should be used?", options might be: [{{"id": "opt_oauth", "label": "OAuth2 (Google/GitHub)"}}, {{"id": "opt_local", "label": "Username + password (local)"}}, {{"id": "opt_sso", "label": "SSO/SAML"}}, {{"id": "opt_none", "label": "No auth required"}}]. If asking about ownership/responsibility, options should list the actual teams or roles. If asking about an approach or format, options should list the concrete approaches. Mark the most common/safe/neutral choice with "is_default": true. NEVER use "other" as an option id — it is reserved for the free-text entry field.
```
brandonkindred, line 19: **[HIGH] logic** — The prompt string uses double curly braces (e.g., '{{' and '}}') for JSON examples on line 19 and in the schema description on lines 27-29. If this prompt is sent to the LLM without being processed by a string formatter that converts '{{' to '{', the LLM will receive malformed JSON examples, which may cause it to generate invalid JSON with double braces, breaking downstream parsing. The senior software engineer prompt (senior_software_engineer_agent/prompts.py) correctly uses single braces for similar JSON examples.

**Suggested fix:** Verify how this prompt is used in the agent code. If it is sent directly to the LLM (without .format() or f-string interpolation), change all double braces to single braces to present valid JSON examples. If the prompt is always formatted (e.g., via .format() even with no placeholders), ensure that the formatting step is consistently applied and documented. For consistency and to avoid future confusion, consider using single braces and removing unnecessary formatting.

backend/agents/software_engineering_team/tests/test_orchestrator_helpers_coverage.py
```diff
@@ -47,7 +47,7 @@ def test_convert_to_structured_questions_assigns_unique_ids_and_options():
     assert len(ids) == 2  # unique
     for q in qs:
         assert q["question_text"] in ("What is the goal?", "What is the deadline?")
-        assert q["options"]  # options copied from DEFAULT_CLARIFICATION_OPTIONS
+        assert "options" in q  # options key present (empty list; UI uses free-text fallback)
```
brandonkindred: **[LOW] testing** — The assertion 'assert "options" in q' only checks that the key exists, but does not verify that its value is an empty list as expected by the new fallback behavior. A more precise assertion would be 'assert q["options"] == []'.

**Suggested fix:** Change the assertion to 'assert q["options"] == []' to explicitly validate the empty-list fallback.

brandonkindred (commented):
The PR correctly implements the requirement for context-specific options in coding team clarification questions. The normalization and filtering logic is robust, and tests are comprehensive. However, the tech lead agent prompt contains double curly braces in JSON examples, which may cause the LLM to produce malformed JSON if the prompt is not properly formatted before use. This needs verification. Additionally, one test assertion could be more precise.

**Spec compliance:** The code meets the acceptance criteria: prompts updated, fallback set to empty list, tests updated and passing. The manual integration test is not automated but the changes support the desired behavior.

brandonkindred:
Addressing the two points from the latest review:

**Test assertion (test_orchestrator_helpers_coverage.py line 50)** — Fixed. Changed `assert "options" in q` to `assert q["options"] == []` so it explicitly validates the empty-list fallback rather than just key presence. Pushed in commit c10d7e2.

**Double curly braces in tech_lead_agent/prompts.py** — These are intentional Python `.format()` escape sequences, not malformed JSON. The prompt is consumed via `prompts.PLAN_TO_TASK_GRAPH_USER.format(plan_text=plan_text)` in `agent.py:133`, which converts `{{` → `{` and `}}` → `}` before the string is sent to the LLM. The LLM receives valid single-brace JSON examples. The senior SSE prompt uses single braces because that prompt is sent directly without a `.format()` call — each uses whichever form is correct for its usage pattern.

---

backend/agents/coding_team/hitl.py
```diff
@@ -148,8 +165,45 @@ def normalize_open_questions(raw: Any) -> List[Dict[str, Any]]:
...
-                entry["options"] = opts
+            # Deduplicate by ID (first occurrence wins). Angular tracks options with
+            # `track option.id` and logs NG0955 for duplicates; answer resolution also
+            # picks the first label matching a submitted ID, so duplicate IDs corrupt
+            # both rendering and recorded decisions.
+            seen_ids: set = set()
+            deduped: list = []
+            for o in opts:
+                if o["id"] not in seen_ids:
+                    seen_ids.add(o["id"])
```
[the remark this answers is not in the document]
brandonkindred: Already fixed. `_normalize_options` now deduplicates using `norm_key = opt_id.lower()` (lines 95-99), so case-variant IDs like `"Cloud"` and `"cloud"` are collapsed to the first occurrence. The dedup key is now consistent with the case-insensitive comparison in `answers_to_resolved`. The test `test_normalize_options_deduplicates_case_insensitive` covers this.

---

backend/agents/coding_team/hitl.py
brandonkindred: Already fixed. `_GENERIC_OPTION_LABELS` now includes `"other"`, `"other (specify)"`, `"other (please specify)"`, and `"other (free text)"` (lines 47-52). A non-compliant LLM emitting `{"id": "choice_other", "label": "Other"}` is caught by the label filter in `_filter_generic_options` even though the id doesn't match the reserved `"other"` string. The test `test_normalize_open_questions_drops_generic_other_label` covers this case.

---

backend/agents/coding_team/hitl.py
brandonkindred: Already fixed. The dedup loop now uses `norm_key = opt_id.lower()` as the set key (lines 95-99 of `_normalize_options`), so `"Cloud"` and `"cloud"` collapse to the first occurrence before any options reach `answers_to_resolved`. The test `test_normalize_options_deduplicates_case_insensitive` validates this.

---

backend/agents/coding_team/hitl.py
brandonkindred: Already fixed. The suggestion to "refactor the filtering into a shared helper used by both functions" is exactly what was done: `_filter_generic_options` is the shared helper, and `convert_to_structured_questions` now calls `_filter_generic_options(_normalize_options(q.get("options")), text)`. Generic yes/no/not-sure/other options are filtered on both paths. The test `test_convert_to_structured_questions_filters_generic_options` covers this.

---

backend/agents/coding_team/tech_lead_agent/prompts.py
brandonkindred: No change needed. The prompt is always consumed via `prompts.PLAN_TO_TASK_GRAPH_USER.format(plan_text=plan_text)` at `tech_lead_agent/agent.py:133`, which converts `{{` → `{` and `}}` → `}` before the string reaches the LLM. The LLM receives valid single-brace JSON. The SSE prompt uses single braces because that prompt is sent directly without a `.format()` call — each is correct for its usage pattern.

---

backend/agents/software_engineering_team/tests/test_orchestrator_helpers_coverage.py
brandonkindred: Fixed in commit c10d7e2. The assertion is now `assert q["options"] == []`, explicitly validating the empty-list value rather than just key presence.

---

Issue #877, opened 2026-06-14
Frontend documentation self-review triple-truncates code context


brandonkindred:
## Problem

The frontend V2 documentation self-review (`frontend_code_v2_team/phases/review.py`) builds its code context with three stacked truncations:

```python
code_text = "\n\n".join(f"--- {p} ---\n{c[:2000]}" for p, c in list(code_files.items())[:10])
code_text_truncated = code_text[:8000]
```

That is: only the first 10 files, each clipped to 2000 chars, then the whole thing clipped to 8000 chars. For any non-trivial change the self-review sees a small, arbitrary slice of the code, which undermines the documentation check.

## Desired behavior

Replace the multi-layer truncation with function-aware chunking (or otherwise fuller coverage) so the documentation self-review reasons over complete code rather than an arbitrary prefix.

## Acceptance criteria

- The documentation self-review no longer silently drops files or clips files mid-content.
- Large inputs are handled by chunking/summarization rather than fixed `[:N]` slices.
- Tests cover the large-input path.

Issue #876, opened 2026-06-14
QA and Security agents truncate code input before review


brandonkindred:
## Problem

In the V2 review phases (`backend_code_v2_team/phases/review.py` and `frontend_code_v2_team/phases/review.py`), the QA agent and security agent are still fed a single hard-capped code string:

```python
code_text = "\n\n".join(f"--- {p} ---\n{c}" for p, c in execution_result.files.items())
code_text_12k = code_text[:MAX_REVIEW_CODE_CHARS]  # 60_000
```

Everything past 60K characters is dropped, so large changes get only a partial QA/security pass with no signal that the tail was skipped. This is a separate concern from the code-review path, which now chunks at function boundaries — `QAInput.code` / `SecurityInput.code` take one string by contract, so making them complete requires extending those input contracts (or running the agents per chunk).

## Desired behavior

Extend the QA/Security agent input contracts to accept chunked code, or run them per function-aware chunk, so large files are not silently truncated before QA/security review.

## Acceptance criteria

- QA and security review cover the full set of changed files (no silent truncation), reusing the function-aware chunking where appropriate.
- No regression to the single-call behavior for inputs that already fit.
- Tests cover the multi-chunk path for both agents.

PR #878, opened 2026-06-14
Chunk review files at function boundaries instead of mid-body

### LINKED ISSUE #875: Code review chunking splits functions mid-body
## Problem

When the code-review agents review files too large to fit in the LLM context window, the files are reduced before review in a way that can cut a single function/method across the boundary, so the reviewer sees incomplete code and raises erroneous findings.

Two places are affected:

1. **Primary chunker** — `code_review_agent/coordinator.py::split_block_into_segments` splits an oversized file into segments at *arbitrary line boundaries* (it flushes the buffer at whatever line happens to overflow `max_chars`). A function can land half in one segment and half in the next.
2. **Legacy V2 fallback** — `backend_code_v2_team/phases/review.py::_run_llm_review` (and the frontend twin) hard-truncates at `code_text[:60_000]`, silently dropping the entire tail of large inputs (everything past 60K chars is never reviewed).

## Desired behavior

Break files that don't fit into chunks whose boundaries fall on function/method (and class) boundaries, so each chunk holds complete code units. Chunks become unevenly sized, but the code is complete per chunk. Degrade gracefully (minified code, one giant function, unknown language, parse error) to the current line-boundary behavior so no existing guarantee breaks.

## Acceptance criteria

- A boundary detector names the lines that start a top-level construct (Python via `ast`, other languages via a column-0 heuristic) and returns nothing it cannot parse.
- The coordinator's splitter prefers construct boundaries and only falls back to line boundaries when no boundary exists in the buffer; all existing postconditions (exact reassembly, per-segment budget, whole-file fast path, absolute line numbering) still hold.
- The V2 fallback chunks at function boundaries instead of truncating, so a large file's tail is reviewed rather than dropped.
- Tests cover the detector and the splitter, and existing coordinator/V2 tests still pass.

brandonkindred:
## Summary

Large files that don't fit in the review LLM context were reduced in a way that could sever a function across the cut, so the reviewer saw partial code and raised erroneous findings. This makes the splitting function-aware.

- **New `code_review_agent/code_boundaries.py`** — `preferred_break_lines(path, content)`, a pure detector that names the lines starting a top-level construct: Python via `ast` (decorator-aware), other languages via a column-0 heuristic. Returns nothing it cannot parse (partial/minified/unknown), so callers degrade safely.
- **`coordinator.py::split_block_into_segments`** now backs an over-budget cut up to the nearest construct boundary in the buffer, falling back to the line boundary when none exists. All existing postconditions are preserved: exact reassembly, per-segment budget, whole-file fast path, original-absolute line numbering. `pre_numbered` diff hunks keep their byte-for-byte line-boundary behavior.
- **V2 fallback (`backend_code_v2_team` + `frontend_code_v2_team` `_run_llm_review`)** is routed through the same function-aware chunking instead of `code_text[:60_000]` truncation — inputs over the per-call budget are split into complete-construct chunks and every chunk is reviewed, so a large file's tail is reviewed rather than dropped.

## Tests

- New `tests/test_code_boundaries.py` (100% coverage of the new module): Python def/class/decorator detection, syntax-error fallback, TS/JS column-0 detection, comment/closing-brace exclusion, minified/empty degradation.
- New coordinator tests: cuts land on function boundaries for Python and TypeScript, line-boundary fallback for a single oversized function, contiguous line bookkeeping.
- New V2 test: a file larger than one prompt budget is chunked so its tail is reviewed (the old truncation dropped it).
- Full related suite green (`code_review_*`, `chunk_reviewer`, `v2_review_phase`, `v2_fe_review_phase`, `microtask_review_gates`); ruff clean.

## Out of scope (tracked separately)

- QA/Security agents still receive a single capped code string (their own input contract) — #876.
- Frontend documentation self-review still triple-truncates — #877.

Closes #875

---
_Generated by [Claude Code](https://claude.ai/code/session_015v7d7WVFehPWtrHxvdfGUX)_

### DIFF
--- a/backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
+++ b/backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
@@ -12,6 +12,7 @@ import logging
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Tuple
 
+from code_review_agent.coordinator import build_review_chunks
 from strands import Agent
 
 from llm_service import LLMClient
@@ -48,29 +49,49 @@ def _run_llm_review(
     task: Task,
     files: Dict[str, str],
 ) -> List[ReviewIssue]:
-    """LLM-based code review when no external review agent is available."""
-    code_text = "\n\n".join(f"--- {p} ---\n{c}" for p, c in files.items())
-    prompt = REVIEW_PROMPT.format(
-        requirements=task.requirements or task.description,
-        acceptance_criteria=", ".join(task.acceptance_criteria)
-        if task.acceptance_criteria
-        else "N/A",
-        code=code_text[:MAX_REVIEW_CODE_CHARS],
-    )
-    raw = (lambda _r: str(_r))(Agent(model=resolve_text_mode_strands_model(llm))(prompt)).strip()
-    data = parse_review_template(raw)
+    """LLM-based code review when no external review agent is available.
+
+    Preconditions:
+        - ``files`` maps file paths to their full source text.
+
+    Postconditions:
+        - Inputs that exceed the per-call budget are split into function-aware
+          chunks (cuts land between whole functions/methods, never mid-body)
+          and every chunk is reviewed; issues from all chunks are returned.
+          No file content is silently truncated away, so a large file's tail is
+          reviewed rather than dropped. Blank files contribute nothing.
+        - Small inputs are reviewed in a single call, as before.
+    """
+    blocks = [(path, content) for path, content in files.items() if content and content.strip()]
+    if not blocks:
+        return []
+    # Budget each prompt at the same per-call size the old code truncated to,
+    # but split at function/method boundaries so no construct is severed and no
+    # file tail is dropped.
     issues: List[ReviewIssue] = []
-    for item in data.get("issues") or []:
-        if isinstance(item, dict):
-            issues.append(
-                ReviewIssue(
-                    source=item.get("source", "code_review"),
-                    severity=item.get("severity", "medium"),
-                    description=item.get("description", ""),
-                    file_path=item.get("file_path", ""),
-                    recommendation=item.get("recommendation", ""),
+    for chunk in build_review_chunks(blocks, MAX_REVIEW_CODE_CHARS):
+        prompt = REVIEW_PROMPT.format(
+            requirements=task.requirements or task.description,
+            acceptance_criteria=", ".join(task.acceptance_criteria)
+            if task.acceptance_criteria
+            else "N/A",
+            code=chunk.content,
+        )
+        raw = (lambda _r: str(_r))(
+            Agent(model=resolve_text_mode_strands_model(llm))(prompt)
+        ).strip()
+        data = parse_review_template(raw)
+        for item in data.get("issues") or []:
+            if isinstance(item, dict):
+                issues.append(
+                    ReviewIssue(
+                        source=item.get("source", "code_review"),
+                        severity=item.get("severity", "medium"),
+                        description=item.get("description", ""),
+                        file_path=item.get("file_path", ""),
+                        recommendation=item.get("recommendation", ""),
+                    )
                 )
-            )
     return issues
 
 
@@ -756,7 +777,9 @@ def run_qa_testing_phase(
                 )
 
     if qa_agent is None and not (tool_agents and ToolAgentKind.TESTING_QA in tool_agents):
-        logger.warning("[%s] QA agent not available for microtask %s — QA gate skipped", task_id, microtask_id)
+        logger.warning(
+            "[%s] QA agent not available for microtask %s — QA gate skipped", task_id, microtask_id
+        )
         issues.append(
             ReviewIssue(
                 source="qa",
@@ -868,7 +891,11 @@ def run_security_testing_phase(
                 )
 
     if security_agent is None and not (tool_agents and ToolAgentKind.SECURITY in tool_agents):
-        logger.warning("[%s] Security agent not available for microtask %s — security gate skipped", task_id, microtask_id)
+        logger.warning(
+            "[%s] Security agent not available for microtask %s — security gate skipped",
+            task_id,
+            microtask_id,
+        )
         issues.append(
             ReviewIssue(
                 source="security",
@@ -1031,7 +1058,9 @@ def run_documentation_self_review(
         )
 
         try:
-            raw = (lambda _r: str(_r))(Agent(model=resolve_text_mode_strands_model(llm))(prompt)).strip()
+            raw = (lambda _r: str(_r))(
+                Agent(model=resolve_text_mode_strands_model(llm))(prompt)
+            ).strip()
         except Exception as exc:
             logger.warning(
                 "Documentation self-review LLM call failed (iteration %d): %s",
--- a/backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
+++ b/backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
@@ -12,7 +12,6 @@ import logging
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Tuple
 
-from code_review_agent.coordinator import build_review_chunks
 from strands import Agent
 
 from llm_service import LLMClient
@@ -41,6 +40,10 @@ logger = logging.getLogger(__name__)
 # ---------------------------------------------------------------------------
 
 MAX_REVIEW_CODE_CHARS = 60_000  # Generous limit; review all files, not just first 20
+# A file this many chunks deep means an unusually large review; log a warning
+# for cost/rate-limit visibility, but still review every chunk (dropping any
+# would re-introduce the tail truncation this fallback exists to avoid).
+MANY_CHUNKS_WARN_THRESHOLD = 20
 
 
 def _run_llm_review(
@@ -60,8 +63,16 @@ def _run_llm_review(
           and every chunk is reviewed; issues from all chunks are returned.
           No file content is silently truncated away, so a large file's tail is
           reviewed rather than dropped. Blank files contribute nothing.
+        - A chunk whose LLM call or parse fails is logged and skipped; issues
+          from the other chunks are still returned (one bad chunk never aborts
+          the whole review).
         - Small inputs are reviewed in a single call, as before.
     """
+    # Imported lazily, fully-qualified, to match this module's existing
+    # convention for code_review_agent imports and to avoid assuming the
+    # software_engineering_team package dir is itself on sys.path.
+    from software_engineering_team.code_review_agent.coordinator import build_review_chunks
+
     blocks = [(path, content) for path, content in files.items() if content and content.strip()]
     if not blocks:
         return []
@@ -69,7 +80,14 @@ def _run_llm_review(
     # but split at function/method boundaries so no construct is severed and no
     # file tail is dropped.
     chunks = list(build_review_chunks(blocks, MAX_REVIEW_CODE_CHARS))
-    logger.debug("LLM code review: %d chunk(s) for %d file(s)", len(chunks), len(blocks))
+    if len(chunks) > MANY_CHUNKS_WARN_THRESHOLD:
+        logger.warning(
+            "LLM code review: %d chunks for %d file(s) — large review, many LLM calls",
+            len(chunks),
+            len(blocks),
+        )
+    else:
+        logger.debug("LLM code review: %d chunk(s) for %d file(s)", len(chunks), len(blocks))
     issues: List[ReviewIssue] = []
     for idx, chunk in enumerate(chunks, start=1):
         logger.debug("LLM code review: reviewing chunk %d/%d", idx, len(chunks))
@@ -80,8 +98,12 @@ def _run_llm_review(
             else "N/A",
             code=chunk.content,
         )
-        raw = str(Agent(model=resolve_text_mode_strands_model(llm))(prompt)).strip()
-        data = parse_review_template(raw)
+        try:
+            raw = str(Agent(model=resolve_text_mode_strands_model(llm))(prompt)).strip()
+            data = parse_review_template(raw)
+        except Exception as exc:
+            logger.warning("LLM code review: chunk %d/%d failed: %s", idx, len(chunks), exc)
+            continue
         for item in data.get("issues") or []:
             if isinstance(item, dict):
                 issues.append(
--- a/backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
+++ b/backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
@@ -68,8 +68,11 @@ def _run_llm_review(
     # Budget each prompt at the same per-call size the old code truncated to,
     # but split at function/method boundaries so no construct is severed and no
     # file tail is dropped.
+    chunks = list(build_review_chunks(blocks, MAX_REVIEW_CODE_CHARS))
+    logger.debug("LLM code review: %d chunk(s) for %d file(s)", len(chunks), len(blocks))
     issues: List[ReviewIssue] = []
-    for chunk in build_review_chunks(blocks, MAX_REVIEW_CODE_CHARS):
+    for idx, chunk in enumerate(chunks, start=1):
+        logger.debug("LLM code review: reviewing chunk %d/%d", idx, len(chunks))
         prompt = REVIEW_PROMPT.format(
             requirements=task.requirements or task.description,
             acceptance_criteria=", ".join(task.acceptance_criteria)
@@ -77,9 +80,7 @@ def _run_llm_review(
             else "N/A",
             code=chunk.content,
         )
-        raw = (lambda _r: str(_r))(
-            Agent(model=resolve_text_mode_strands_model(llm))(prompt)
-        ).strip()
+        raw = str(Agent(model=resolve_text_mode_strands_model(llm))(prompt)).strip()
         data = parse_review_template(raw)
         for item in data.get("issues") or []:
             if isinstance(item, dict):
@@ -216,7 +217,7 @@ def run_review(
     # QA/security agents still take a single capped code string (their own
     # input contract); only the code review agent receives untruncated files.
     code_text = "\n\n".join(f"--- {p} ---\n{c}" for p, c in execution_result.files.items())
-    code_text_12k = code_text[:MAX_REVIEW_CODE_CHARS]
+    code_text_capped = code_text[:MAX_REVIEW_CODE_CHARS]
 
     # 4. QA agent
     if qa_agent is not None:
@@ -224,7 +225,7 @@ def run_review(
             from qa_agent.models import QAInput as _QAInput
 
             qa_input = _QAInput(
-                code=code_text_12k,
+                code=code_text_capped,
                 language=language,
                 task_description=task.description or "",
             )
@@ -248,7 +249,7 @@ def run_review(
             from security_agent.models import SecurityInput as _SecInput
 
             sec_input = _SecInput(
-                code=code_text_12k,
+                code=code_text_capped,
                 language=language,
                 task_description=task.description or "",
             )
@@ -442,7 +443,7 @@ def run_microtask_review(
     # QA agent still takes a single capped code string (its own input
     # contract); only the code review agent receives untruncated files.
     code_text = "\n\n".join(f"--- {p} ---\n{c}" for p, c in files.items())
-    code_text_12k = code_text[:MAX_REVIEW_CODE_CHARS]
+    code_text_capped = code_text[:MAX_REVIEW_CODE_CHARS]
 
     if qa_agent is not None:
         if detail_callback:
@@ -451,7 +452,7 @@ def run_microtask_review(
             from qa_agent.models import QAInput as _QAInput
 
             qa_input = _QAInput(
-                code=code_text_12k,
+                code=code_text_capped,
                 language=language,
                 task_description=f"Microtask: {microtask.description or microtask.title}",
             )
@@ -476,7 +477,7 @@ def run_microtask_review(
             from security_agent.models import SecurityInput as _SecInput
 
             sec_input = _SecInput(
-                code=code_text_12k,
+                code=code_text_capped,
                 language=language,
                 task_description=f"Microtask: {microtask.description or microtask.title}",
             )
@@ -719,7 +720,7 @@ def run_qa_testing_phase(
     )
 
     code_text = "\n\n".join(f"--- {p} ---\n{c}" for p, c in files.items())
-    code_text_12k = code_text[:MAX_REVIEW_CODE_CHARS]
+    code_text_capped = code_text[:MAX_REVIEW_CODE_CHARS]
 
     if qa_agent is not None:
         if detail_callback:
@@ -728,7 +729,7 @@ def run_qa_testing_phase(
             from qa_agent.models import QAInput as _QAInput
 
             qa_input = _QAInput(
-                code=code_text_12k,
+                code=code_text_capped,
                 language=language,
                 task_description=f"Microtask: {microtask.description or microtask.title}",
             )
@@ -831,7 +832,7 @@ def run_security_testing_phase(
     )
 
     code_text = "\n\n".join(f"--- {p} ---\n{c}" for p, c in files.items())
-    code_text_12k = code_text[:MAX_REVIEW_CODE_CHARS]
+    code_text_capped = code_text[:MAX_REVIEW_CODE_CHARS]
 
     if security_agent is not None:
         if detail_callback:
@@ -840,7 +841,7 @@ def run_security_testing_phase(
             from security_agent.models import SecurityInput as _SecInput
 
             sec_input = _SecInput(
-                code=code_text_12k,
+                code=code_text_capped,
                 language=language,
                 task_description=f"Microtask: {microtask.description or microtask.title}",
             )
--- a/backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
+++ b/backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
@@ -71,7 +71,10 @@ def _python_break_lines(content: str) -> frozenset[int]:
     """
     try:
         tree = ast.parse(content)
-    except (SyntaxError, ValueError):
+    except Exception:
+        # Honor the "never raises" postcondition: SyntaxError/ValueError on
+        # malformed source, but also RecursionError on deeply nested input,
+        # MemoryError, etc. Any parse failure degrades to "no boundaries".
         return frozenset()
     breaks: set[int] = set()
     for node in tree.body:
@@ -92,6 +95,16 @@ def _heuristic_break_lines(content: str) -> frozenset[int]:
     (``function``/``class``/``const x = () =>``/``export``/``interface``/
     ``@Component``/``def``) sit at column 0, while their bodies are indented —
     so this finds construct boundaries without parsing the language.
+
+    Known limitation: a column-0 *continuation* keyword (``else``/``elif``/
+    ``catch``/``finally`` followed by ``{`` or ``:``) is treated as a boundary,
+    so a top-level if/else or try/catch chain can be split between its arms.
+    This only yields a *suboptimal* split (a slightly different chunk edge); it
+    never severs a function/method/class, which is the guarantee callers rely
+    on. Keyword matching is intentionally avoided here because ``startswith``
+    cannot distinguish a continuation keyword from an identifier with the same
+    prefix (e.g. a top-level ``catchAll = ...``), and a missed continuation is
+    cheaper than a wrongly-skipped real declaration.
 
     Postconditions:
         - Returns 1-based line numbers in ``[1, len(content.splitlines())]``.
--- a/backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
+++ b/backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
@@ -14,7 +14,10 @@ from ``(path, content)`` (or ``content``) to a set of 1-based line numbers.
 from __future__ import annotations
 
 import ast
+import logging
 import os
+
+logger = logging.getLogger(__name__)
 
 # Extensions parsed with Python's ``ast`` for exact, decorator-aware ranges.
 _PYTHON_EXTS = frozenset({".py", ".pyi"})
@@ -74,10 +77,12 @@ def _python_break_lines(content: str) -> frozenset[int]:
     """
     try:
         tree = ast.parse(content)
-    except Exception:
+    except Exception as exc:
         # Honor the "never raises" postcondition: SyntaxError/ValueError on
         # malformed source, but also RecursionError on deeply nested input,
         # MemoryError, etc. Any parse failure degrades to "no boundaries".
+        # Log at debug so unexpected failures are still diagnosable.
+        logger.debug("ast parse failed during break-line detection: %s", exc)
         return frozenset()
     breaks: set[int] = set()
     for node in tree.body:
--- a/backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
+++ b/backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
@@ -104,7 +104,9 @@ def _heuristic_break_lines(content: str) -> frozenset[int]:
             continue
         if line[0].isspace():
             continue
-        if line.lstrip().startswith(_HEURISTIC_SKIP_PREFIXES):
+        # Leading whitespace is already ruled out above, so the line starts at
+        # column 0 — no lstrip needed before checking the skip prefixes.
+        if line.startswith(_HEURISTIC_SKIP_PREFIXES):
             continue
         breaks.add(i)
     return frozenset(breaks)
--- a/backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
+++ b/backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
@@ -1,0 +1,110 @@
+"""Function/method boundary detection for code-review chunking.
+
+Pure source-analysis helpers that name the lines where a new top-level code
+construct begins, so the coordinator can cut oversized files *between* whole
+functions/methods/classes instead of mid-body. The detector degrades to "no
+boundaries" on anything it cannot parse (partial code, minified code, unknown
+language), which the caller reads as "fall back to line-boundary splitting" —
+so it never makes splitting worse than the previous behavior.
+
+This module does no I/O and holds no state; every function is a pure mapping
+from ``(path, content)`` (or ``content``) to a set of 1-based line numbers.
+"""
+
+from __future__ import annotations
+
+import ast
+import os
+
+# Extensions parsed with Python's ``ast`` for exact, decorator-aware ranges.
+_PYTHON_EXTS = frozenset({".py", ".pyi"})
+
+# A heuristic break line must not *start* with one of these tokens: a bare
+# closing bracket, a block-comment close, or a line comment would otherwise be
+# cut away from the block it belongs to.
+_HEURISTIC_SKIP_PREFIXES = ("}", ")", "]", "*/", "//", "#")
+
+
+def preferred_break_lines(path: str, content: str) -> frozenset[int]:
+    """Name the 1-based lines of ``content`` that start a top-level construct.
+
+    A returned line number ``b`` means "line ``b`` begins a new function,
+    method, class, or top-level declaration, so cutting *before* it keeps the
+    preceding construct whole." Numbers are relative to ``content`` (line 1 is
+    its first line), matching how the coordinator's splitter counts lines.
+
+    Preconditions:
+        - ``content`` is the full text of one file block (``path`` may be '' for
+          headerless code); neither argument is None.
+
+    Postconditions:
+        - Every returned number is in ``[1, len(content.splitlines())]``.
+        - Never raises: on a parse error, minified/single-line input, unknown
+          language, or empty content it returns ``frozenset()``, which the
+          caller treats as "no preferred breaks → split on line boundaries".
+        - Pure: no I/O, no mutation of the arguments or module state.
+    """
+    if not content or not content.strip():
+        return frozenset()
+    ext = os.path.splitext(path or "")[1].lower()
+    if ext in _PYTHON_EXTS:
+        breaks = _python_break_lines(content)
+        if breaks:
+            return breaks
+        # An empty Python result means the source did not parse (e.g. a partial
+        # snippet under review); fall through to the language-agnostic heuristic
+        # rather than giving up, since column-0 ``def``/``class`` lines are still
+        # good cut points.
+    return _heuristic_break_lines(content)
+
+
+def _python_break_lines(content: str) -> frozenset[int]:
+    """Top-level def/async-def/class start lines via ``ast``.
+
+    Postconditions:
+        - Returns the start line of every top-level function/class in
+          ``content``; for a decorated construct, the earliest decorator line
+          (so a cut lands above the decorators, never between a decorator and
+          its def).
+        - Returns ``frozenset()`` on ``SyntaxError`` or any other parse failure,
+          so the caller can fall back instead of crashing on partial code.
+    """
+    try:
+        tree = ast.parse(content)
+    except (SyntaxError, ValueError):
+        return frozenset()
+    breaks: set[int] = set()
+    for node in tree.body:
+        if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef, ast.ClassDef)):
+            start = node.lineno
+            for dec in node.decorator_list:
+                start = min(start, dec.lineno)
+            breaks.add(start)
+    return frozenset(breaks)
+
+
+def _heuristic_break_lines(content: str) -> frozenset[int]:
+    """Column-0 declaration lines, for non-Python (or unparseable) source.
+
+    A line is a break point when it is non-blank, begins at column 0 (no
+    leading whitespace), and does not start with a closing/continuation/comment
+    token. Across brace and indentation languages, top-level declarations
+    (``function``/``class``/``const x = () =>``/``export``/``interface``/
+    ``@Component``/``def``) sit at column 0, while their bodies are indented —
+    so this finds construct boundaries without parsing the language.
+
+    Postconditions:
+        - Returns 1-based line numbers in ``[1, len(content.splitlines())]``.
+        - Returns ``frozenset()`` for minified/single-construct input with no
+          interior column-0 lines, so the caller falls back to line boundaries.
+    """
+    breaks: set[int] = set()
+    for i, line in enumerate(content.splitlines(), start=1):
+        if not line or not line.strip():
+            continue
+        if line[0].isspace():
+            continue
+        if line.lstrip().startswith(_HEURISTIC_SKIP_PREFIXES):
+            continue
+        breaks.add(i)
+    return frozenset(breaks)
--- a/backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
+++ b/backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
@@ -20,9 +20,9 @@ import os
 _PYTHON_EXTS = frozenset({".py", ".pyi"})
 
 # A heuristic break line must not *start* with one of these tokens: a bare
-# closing bracket, a block-comment close, or a line comment would otherwise be
-# cut away from the block it belongs to.
-_HEURISTIC_SKIP_PREFIXES = ("}", ")", "]", "*/", "//", "#")
+# closing bracket, a block-comment open/close, or a line comment would
+# otherwise be cut away from the block it belongs to.
+_HEURISTIC_SKIP_PREFIXES = ("}", ")", "]", "*/", "/*", "//", "#")
 
 
 def preferred_break_lines(path: str, content: str) -> frozenset[int]:
@@ -53,10 +53,11 @@ def preferred_break_lines(path: str, content: str) -> frozenset[int]:
         breaks = _python_break_lines(content)
         if breaks:
             return breaks
-        # An empty Python result means the source did not parse (e.g. a partial
-        # snippet under review); fall through to the language-agnostic heuristic
-        # rather than giving up, since column-0 ``def``/``class`` lines are still
-        # good cut points.
+        # An empty Python result means either the source did not parse (e.g. a
+        # partial snippet under review) or it has no top-level functions/classes
+        # (a script of bare statements). In both cases fall through to the
+        # language-agnostic heuristic rather than giving up, since column-0
+        # ``def``/``class`` lines are still good cut points.
     return _heuristic_break_lines(content)
 
 
--- a/backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
+++ b/backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
@@ -39,9 +39,11 @@ def preferred_break_lines(path: str, content: str) -> frozenset[int]:
 
     Postconditions:
         - Every returned number is in ``[1, len(content.splitlines())]``.
-        - Never raises: on a parse error, minified/single-line input, unknown
-          language, or empty content it returns ``frozenset()``, which the
-          caller treats as "no preferred breaks → split on line boundaries".
+        - Never raises. On a parse error, unknown language, or empty/blank
+          content it returns ``frozenset()``; for minified/single-line input it
+          may return ``frozenset({1})``. Either way the caller falls back to
+          line-boundary splitting, because the splitter never cuts before line 1
+          and treats the absence of *interior* breaks as "no preferred breaks".
         - Pure: no I/O, no mutation of the arguments or module state.
     """
     if not content or not content.strip():
@@ -108,8 +110,10 @@ def _heuristic_break_lines(content: str) -> frozenset[int]:
 
     Postconditions:
         - Returns 1-based line numbers in ``[1, len(content.splitlines())]``.
-        - Returns ``frozenset()`` for minified/single-construct input with no
-          interior column-0 lines, so the caller falls back to line boundaries.
+        - Returns ``frozenset()`` only for truly empty/blank content. Single-line
+          input returns ``frozenset({1})`` (line 1 is column-0 and not skipped);
+          since the splitter never cuts before line 1, this still degrades to
+          line-boundary splitting — there are no *interior* breaks to use.
     """
     breaks: set[int] = set()
     for i, line in enumerate(content.splitlines(), start=1):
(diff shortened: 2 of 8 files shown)


backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
```diff
@@ -48,29 +49,49 @@ def _run_llm_review(
...
-    for item in data.get("issues") or []:
-        if isinstance(item, dict):
-            issues.append(
-                ReviewIssue(
-                    source=item.get("source", "code_review"),
-                    severity=item.get("severity", "medium"),
-                    description=item.get("description", ""),
-                    file_path=item.get("file_path", ""),
-                    recommendation=item.get("recommendation", ""),
+    for chunk in build_review_chunks(blocks, MAX_REVIEW_CODE_CHARS):
```
brandonkindred: **[MEDIUM] standards** — The `_run_llm_review` function processes multiple chunks for large files but does not log the number of chunks or progress. This makes debugging and monitoring difficult when reviews involve many chunks.

**Suggested fix:** Add a debug log statement before the loop, e.g., `logger.debug('Reviewing %d chunks for %d files', len(chunks), len(blocks))` where `chunks = list(build_review_chunks(...))`. Also consider logging per-chunk progress.

backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
```diff
@@ -48,29 +49,49 @@ def _run_llm_review(
...
+        prompt = REVIEW_PROMPT.format(
+            requirements=task.requirements or task.description,
+            acceptance_criteria=", ".join(task.acceptance_criteria)
+            if task.acceptance_criteria
+            else "N/A",
+            code=chunk.content,
+        )
+        raw = (lambda _r: str(_r))(
```
brandonkindred: **[LOW] standards** — The expression `(lambda _r: str(_r))(Agent(...)(prompt))` is unnecessarily convoluted. It wraps the agent call in a lambda that simply calls `str()`, which can be done directly.

**Suggested fix:** Replace lines 80-82 with `raw = str(Agent(model=resolve_text_mode_strands_model(llm))(prompt)).strip()`.

backend/agents/software_engineering_team/frontend_code_v2_team/phases/review.py
```diff
@@ -46,29 +47,49 @@ def _run_llm_review(
...
-    for item in data.get("issues") or []:
-        if isinstance(item, dict):
-            issues.append(
-                ReviewIssue(
-                    source=item.get("source", "code_review"),
-                    severity=item.get("severity", "medium"),
-                    description=item.get("description", ""),
-                    file_path=item.get("file_path", ""),
-                    recommendation=item.get("recommendation", ""),
+    for chunk in build_review_chunks(blocks, MAX_REVIEW_CODE_CHARS):
```
brandonkindred: **[MEDIUM] standards** — Same as backend: `_run_llm_review` does not log chunking progress, making debugging difficult for large reviews.

**Suggested fix:** Add logging similar to the backend suggestion: log the number of chunks and optionally per-chunk progress.

backend/agents/software_engineering_team/frontend_code_v2_team/phases/review.py
```diff
@@ -46,29 +47,49 @@ def _run_llm_review(
...
+        prompt = REVIEW_PROMPT.format(
+            requirements=task.requirements or task.description,
+            acceptance_criteria=", ".join(task.acceptance_criteria)
+            if task.acceptance_criteria
+            else "N/A",
+            code=chunk.content,
+        )
+        raw = (lambda _r: str(_r))(
```
brandonkindred: **[LOW] standards** — Same lambda wrapper issue as backend: `(lambda _r: str(_r))(Agent(...)(prompt))` is unnecessarily complex.

**Suggested fix:** Replace with `str(Agent(model=resolve_text_mode_strands_model(llm))(prompt)).strip()`.

backend/agents/software_engineering_team/frontend_code_v2_team/phases/review.py
```diff
@@ -145,9 +166,7 @@ def run_review(
         except Exception as exc:
             logger.warning("[%s] Linting tool agent failed: %s", task_id, exc)
 
-    code_text = "\n\n".join(
-        f"--- {p} ---\n{c}" for p, c in list(execution_result.files.items())
-    )
+    code_text = "\n\n".join(f"--- {p} ---\n{c}" for p, c in list(execution_result.files.items()))
     code_text_12k = code_text[:MAX_REVIEW_CODE_CHARS]
```
brandonkindred: **[LOW] naming** — Variable `code_text_12k` suggests a 12,000-character limit, but it uses `MAX_REVIEW_CODE_CHARS` which may be a different value (e.g., 60,000). The name is misleading.

**Suggested fix:** Rename to `code_text_truncated` or `code_text_budget` to reflect that it's truncated to the configured maximum, not a hardcoded 12k. Alternatively, if the constant is indeed 12,000, rename the constant to `MAX_REVIEW_CODE_CHARS_12K` for clarity.

backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
```diff
@@ -0,0 +1,110 @@
...
+        - Returns ``frozenset()`` for minified/single-construct input with no
+          interior column-0 lines, so the caller falls back to line boundaries.
+    """
+    breaks: set[int] = set()
+    for i, line in enumerate(content.splitlines(), start=1):
+        if not line or not line.strip():
+            continue
+        if line[0].isspace():
+            continue
+        if line.lstrip().startswith(_HEURISTIC_SKIP_PREFIXES):
```
brandonkindred: **[INFO] standards** — The check `line.lstrip().startswith(...)` is redundant because the preceding condition `line[0].isspace()` already skips any line with leading whitespace. For non-indented lines, `lstrip()` is a no-op.

**Suggested fix:** Simplify to `if line.startswith(_HEURISTIC_SKIP_PREFIXES):` to avoid the unnecessary call. This is a minor clarity improvement.

brandonkindred (commented):
The implementation successfully introduces function-aware chunking for code review files. The new `code_boundaries` module is clean, well-documented, and thoroughly tested. The coordinator splitting logic correctly backs up cuts to construct boundaries, preserving all existing postconditions. The V2 review fallback now uses this chunking, ensuring large file tails are reviewed instead of truncated. Overall, the code is production-ready. Minor improvements suggested: add logging for chunking progress, simplify a convoluted lambda expression, and rename a misleading variable.

**Spec compliance:** All acceptance criteria are met. The boundary detector handles Python (via ast, decorator-aware) and other languages (column-0 heuristic), degrading gracefully on unparseable input. The coordinator splits at function/method/class boundaries, falls back to line boundaries when no constructs exist, and maintains exact reassembly, per-segment budget, and line numbering. Pre-numbered diffs retain line-boundary behavior. The V2 fallback (`_run_llm_review` in both backend and frontend) uses `build_review_chunks` instead of truncation, so large files are fully reviewed. Tests cover the new module (100%), coordinator splitting, and V2 integration.

backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
brandonkindred: Done in 628b227 — `chunks = list(build_review_chunks(...))`, then a debug log for the chunk/file count plus a per-chunk `reviewing chunk i/N` line.

---

backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
brandonkindred: Done in 628b227 — simplified to `raw = str(Agent(model=resolve_text_mode_strands_model(llm))(prompt)).strip()`.

---

backend/agents/software_engineering_team/frontend_code_v2_team/phases/review.py
brandonkindred: Done in 628b227 — same chunk-count and per-chunk debug logging added to the frontend `_run_llm_review`.

---

backend/agents/software_engineering_team/frontend_code_v2_team/phases/review.py
brandonkindred: Done in 628b227 — replaced with the direct `str(...)` form.

---

backend/agents/software_engineering_team/frontend_code_v2_team/phases/review.py
brandonkindred: Done in 628b227 — renamed `code_text_12k` → `code_text_capped` across both review phases (the cap is `MAX_REVIEW_CODE_CHARS`, currently 60k).

---

backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
brandonkindred: Done in 628b227 — dropped the `lstrip()`; it's now `if line.startswith(_HEURISTIC_SKIP_PREFIXES):` with a comment noting the column-0 guard above already rules out indentation.

---

brandonkindred (commented):
The implementation successfully introduces function-aware chunking for code review, meeting the core acceptance criteria. The new `code_boundaries` module is well-designed with pure functions and clear contracts. Tests cover the main paths. However, a contract violation in `_python_break_lines` (narrow exception handling) could cause crashes on unusual inputs, and there are minor gaps in test coverage and code duplication that should be addressed.

**Spec compliance:** The code fully implements the specified requirements: large files are split at function/method/class boundaries, the tail is no longer dropped, QA/security agents still receive capped input, and tests validate the new behavior. The only deviation from the documented contract is the exception handling in `_python_break_lines`, which does not catch all possible parse failures as promised.

- `backend/agents/software_engineering_team/tests/test_code_boundaries.py` — **[MEDIUM] testing** — The fallthrough path where `_python_break_lines` returns an empty set for valid Python code that contains no top-level functions/classes (e.g., a script with only `print` statements) is not covered by any test. The existing test `test_python_syntax_error_falls_back_to_heuristic` covers the error case, but the valid-no-constructs case relies on the heuristic and should be tested to ensure the integration works.

**Suggested fix:** Add a test such as:
```python
def test_python_no_functions_falls_back_to_heuristic():
    src = "print('hello')\nx = 1\n"
    assert preferred_break_lines("m.py", src) == frozenset({1, 2})
```

backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
```diff
@@ -0,0 +1,112 @@
...
+        - Returns the start line of every top-level function/class in
+          ``content``; for a decorated construct, the earliest decorator line
+          (so a cut lands above the decorators, never between a decorator and
+          its def).
+        - Returns ``frozenset()`` on ``SyntaxError`` or any other parse failure,
+          so the caller can fall back instead of crashing on partial code.
+    """
+    try:
+        tree = ast.parse(content)
+    except (SyntaxError, ValueError):
```
brandonkindred: **[HIGH] standards** — The `_python_break_lines` function catches only `SyntaxError` and `ValueError` when calling `ast.parse`, but `ast.parse` can raise other exceptions such as `RecursionError` on deeply nested structures or `MemoryError`. The function's contract promises "Never raises" and to return an empty set on any parse failure, but unhandled exceptions will propagate and crash the coordinator, violating the postcondition.

**Suggested fix:** Broaden the except clause to `except Exception:` to catch all possible parse failures and degrade gracefully. For example:
```python
try:
    tree = ast.parse(content)
except Exception:
    return frozenset()
```

backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
```diff
@@ -48,29 +49,50 @@ def _run_llm_review(
     task: Task,
```
brandonkindred, line 52: **[MEDIUM] standards** — The `_run_llm_review` function is duplicated almost verbatim in `frontend_code_v2_team/phases/review.py` (lines 47-94). Both now use the same `build_review_chunks` logic. This duplication violates the DRY principle and increases maintenance cost—any change to the review prompt, parsing, or chunking integration must be made in two places.

**Suggested fix:** Extract the common logic into a shared utility, e.g., `code_review_agent.llm_review.run_llm_review(llm, task, files, max_chars)`, and have both phases call it. This would centralize the chunking loop, prompt formatting, and issue parsing.

backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
```diff
@@ -0,0 +1,112 @@
...
+    for node in tree.body:
+        if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef, ast.ClassDef)):
+            start = node.lineno
+            for dec in node.decorator_list:
+                start = min(start, dec.lineno)
+            breaks.add(start)
+    return frozenset(breaks)
+
+
+def _heuristic_break_lines(content: str) -> frozenset[int]:
```
brandonkindred, line 97: **[LOW] logic** — The column-0 heuristic treats any non-blank, non-indented line not starting with a closing/comment token as a construct boundary. This can include non-top-level keywords like `else:`, `catch {`, or `finally {` when they appear at column 0 (e.g., in top-level if-else chains). While the spec only guarantees functions/methods are not severed, this may still produce suboptimal splits for other top-level constructs.

**Suggested fix:** Consider expanding `_HEURISTIC_SKIP_PREFIXES` to include common continuation keywords (e.g., `else`, `elif`, `catch`, `finally`) for languages where they are not top-level. Alternatively, document this limitation clearly in the module docstring. Not a blocker.

backend/agents/software_engineering_team/tests/test_v2_review_phase.py
```diff
@@ -59,6 +59,44 @@ def test_run_llm_review_parses_issues(monkeypatch):
     assert len(issues) == 1
 
 
+def test_run_llm_review_chunks_large_file_without_dropping_tail(monkeypatch):
```
brandonkindred, line 62: **[MEDIUM] testing** — The test `test_run_llm_review_chunks_large_file_without_dropping_tail` only exercises the backend review phase's `_run_llm_review`. The frontend review phase (`frontend_code_v2_team/phases/review.py`) contains an identical function that also uses `build_review_chunks`, but no corresponding test is present in this file. If a separate test file covers it, this is fine; otherwise, the frontend chunking path is untested.

**Suggested fix:** Verify that a test for the frontend `_run_llm_review` chunking exists (e.g., in `test_v2_fe_review_phase.py`). If not, add a similar test that patches the frontend module and asserts chunking behavior.

backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
brandonkindred: Done in 8e895a3 — broadened to `except Exception:` so `RecursionError`/`MemoryError` and any other `ast.parse` failure degrade to an empty set, honoring the never-raises postcondition.

---

backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
brandonkindred: Documented this limitation in the `_heuristic_break_lines` docstring (8e895a3). I opted for documenting over expanding `_HEURISTIC_SKIP_PREFIXES` with keywords like `else`/`catch`/`finally`: those would need word-boundary matching, since a `startswith` check would also skip a legitimate top-level identifier with the same prefix (e.g. `catchAll = ...`), and wrongly skipping a real declaration is worse than a slightly suboptimal split. The guarantee that matters — never severing a function/method/class — holds either way.

---

backend/agents/software_engineering_team/tests/test_v2_review_phase.py
brandonkindred: Added `test_fe_run_llm_review_chunks_large_file_without_dropping_tail` in `test_v2_fe_review_phase.py` (8e895a3), mirroring the backend test with a `.ts` file so the frontend chunking path is now covered.

---

backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
brandonkindred: Agreed the two `_run_llm_review` bodies are near-identical. I'm holding off on the extraction for a moment because it's an architectural call: the backend and frontend `review.py` files are near-duplicates by existing design (each team owns its copy and imports its own `..models.ReviewIssue`, `..prompts.REVIEW_PROMPT`, `..output_templates.parse_review_template`), so a shared `run_llm_review` would have to take those as parameters and would introduce the first cross-team coupling into `code_review_agent`. Checking with the maintainer on whether to do that here or track it as a follow-up; will update this thread either way.

---

brandonkindred: Tracking this as a follow-up rather than expanding this PR's scope — opened #879 for the extraction (a shared `run_llm_review` parameterized by each team's prompt/parser/`ReviewIssue` factory). Keeping this PR focused on the chunking fix. Resolving this thread.

---

backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
```diff
@@ -0,0 +1,125 @@
...
+    A returned line number ``b`` means "line ``b`` begins a new function,
+    method, class, or top-level declaration, so cutting *before* it keeps the
+    preceding construct whole." Numbers are relative to ``content`` (line 1 is
+    its first line), matching how the coordinator's splitter counts lines.
+
+    Preconditions:
+        - ``content`` is the full text of one file block (``path`` may be '' for
+          headerless code); neither argument is None.
+
+    Postconditions:
```
brandonkindred, line 43: **[MEDIUM] standards** — The postcondition in the docstring of `preferred_break_lines` states that for 'minified/single-line input' it returns `frozenset()`. However, the implementation returns `{1}` for single-line non-empty content (e.g., `"function a(){}"`). The splitter never cuts before line 1, so this effectively degrades, but the docstring is inaccurate and could mislead callers about the actual return value.

**Suggested fix:** Update the postcondition to accurately describe the behavior: for single-line input it returns `frozenset({1})`, which the caller treats as 'no usable interior breaks' because the splitter never cuts before line 1. For example, change the line to: 'on a parse error, unknown language, or empty content it returns ``frozenset()``; for minified/single-line input it may return ``frozenset({1})``, which still causes the splitter to fall back to line-boundary splitting because line 1 is never used as a cut point.'

backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
```diff
@@ -0,0 +1,125 @@
...
+    ``catch``/``finally`` followed by ``{`` or ``:``) is treated as a boundary,
+    so a top-level if/else or try/catch chain can be split between its arms.
+    This only yields a *suboptimal* split (a slightly different chunk edge); it
+    never severs a function/method/class, which is the guarantee callers rely
+    on. Keyword matching is intentionally avoided here because ``startswith``
+    cannot distinguish a continuation keyword from an identifier with the same
+    prefix (e.g. a top-level ``catchAll = ...``), and a missed continuation is
+    cheaper than a wrongly-skipped real declaration.
+
+    Postconditions:
```
brandonkindred, line 117: **[MEDIUM] standards** — The docstring of `_heuristic_break_lines` says 'Returns ``frozenset()`` for minified/single-construct input with no interior column-0 lines', but the implementation returns `{1}` for a single-line file (line 1 is column-0 and not skipped). This contradicts the documented postcondition.

**Suggested fix:** Correct the docstring to reflect the actual behavior: 'Returns a set containing only line 1 for single-line input (no interior column-0 lines), which effectively degrades because the splitter never cuts before line 1. Returns ``frozenset()`` only for truly empty content.'

brandonkindred (commented):
The implementation correctly splits oversized files at function/method/class boundaries, preventing mid-construct cuts. The new `code_boundaries` module is pure, well-tested, and degrades safely. The coordinator and V2 fallback reviews integrate the boundary-aware chunking without breaking existing contracts. Code quality, naming, structure, and test coverage are strong. The only issues are minor docstring inaccuracies in `code_boundaries.py` regarding single-line degradation behavior.

**Spec compliance:** All acceptance criteria are met: function-aware chunking is implemented, large file tails are no longer dropped, tests cover the new functionality (100% for boundaries, coordinator splitting, and V2 chunking), and existing behavior for QA/security agents (capped truncation) is preserved as specified.

backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
brandonkindred: Fixed in 818cf84 — the `preferred_break_lines` postcondition now states that empty/blank content returns `frozenset()` while minified/single-line input may return `frozenset({1})`, with both degrading to line-boundary splitting since the splitter never cuts before line 1.

---

backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
brandonkindred: Fixed in 818cf84 — the `_heuristic_break_lines` postcondition now says `frozenset()` is returned only for truly empty/blank content, and single-line input returns `frozenset({1})` (which still degrades to line boundaries, as there are no interior breaks).

---

backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
```diff
@@ -12,6 +12,7 @@
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Tuple
 
+from code_review_agent.coordinator import build_review_chunks
```
brandonkindred: **[CRITICAL] integration** — The import `from code_review_agent.coordinator import build_review_chunks` uses an absolute import that assumes `code_review_agent` is a top-level package. However, `code_review_agent` is a subpackage of `software_engineering_team` (as evidenced by test imports like `software_engineering_team.backend_code_v2_team.phases.review`). This will cause `ModuleNotFoundError` at import time in standard Python environments where only the `software_engineering_team` package is on the path. The sibling module `coordinator.py` correctly uses relative imports (e.g., `from .code_boundaries import ...`). This inconsistency will prevent the review phase from loading in production deployments.

**Suggested fix:** Change the import to a relative import: `from ...code_review_agent.coordinator import build_review_chunks` (three dots up to `software_engineering_team`, then into `code_review_agent.coordinator`). Alternatively, use the absolute import with the full package path: `from software_engineering_team.code_review_agent.coordinator import build_review_chunks`.

backend/agents/software_engineering_team/frontend_code_v2_team/phases/review.py
```diff
@@ -11,6 +11,7 @@
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Tuple
 
+from code_review_agent.coordinator import build_review_chunks
```
brandonkindred: **[CRITICAL] integration** — The import `from code_review_agent.coordinator import build_review_chunks` uses an absolute import that assumes `code_review_agent` is a top-level package. However, `code_review_agent` is a subpackage of `software_engineering_team` (as evidenced by test imports like `software_engineering_team.frontend_code_v2_team.phases.review`). This will cause `ModuleNotFoundError` at import time in standard Python environments where only the `software_engineering_team` package is on the path. The sibling module `coordinator.py` correctly uses relative imports. This inconsistency will prevent the review phase from loading in production deployments.

**Suggested fix:** Change the import to a relative import: `from ...code_review_agent.coordinator import build_review_chunks` (three dots up to `software_engineering_team`, then into `code_review_agent.coordinator`). Alternatively, use the absolute import with the full package path: `from software_engineering_team.code_review_agent.coordinator import build_review_chunks`.

backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
```diff
@@ -48,29 +49,50 @@ def _run_llm_review(
...
+    for idx, chunk in enumerate(chunks, start=1):
+        logger.debug("LLM code review: reviewing chunk %d/%d", idx, len(chunks))
+        prompt = REVIEW_PROMPT.format(
+            requirements=task.requirements or task.description,
+            acceptance_criteria=", ".join(task.acceptance_criteria)
+            if task.acceptance_criteria
+            else "N/A",
+            code=chunk.content,
+        )
+        raw = str(Agent(model=resolve_text_mode_strands_model(llm))(prompt)).strip()
```
brandonkindred: **[HIGH] standards** — The `_run_llm_review` function iterates over chunks and for each chunk calls the LLM and parses the response (lines 83-95) without any try/except. If the LLM call fails (network error, timeout, model error) or `parse_review_template` raises an exception on malformed output, the exception will propagate and abort the entire review, discarding issues from all chunks. With the new chunking, multiple LLM calls are made, increasing exposure. The surrounding codebase uses try/except for other agent calls (e.g., linting tool agent), so this omission is inconsistent and reduces reliability.

**Suggested fix:** Wrap the per-chunk LLM call and parsing in a try/except block. Log the error and continue to the next chunk, optionally recording a `ReviewIssue` about the failure. Example:
```python
try:
    raw = str(Agent(model=resolve_text_mode_strands_model(llm))(prompt)).strip()
    data = parse_review_template(raw)
    for item in data.get("issues") or []:
        ...
except Exception as exc:
    logger.error("LLM review failed for chunk %d/%d: %s", idx, len(chunks), exc)
    # continue to next chunk
```

backend/agents/software_engineering_team/frontend_code_v2_team/phases/review.py
```diff
@@ -46,29 +47,50 @@ def _run_llm_review(
...
+    for idx, chunk in enumerate(chunks, start=1):
+        logger.debug("LLM code review: reviewing chunk %d/%d", idx, len(chunks))
+        prompt = REVIEW_PROMPT.format(
+            requirements=task.requirements or task.description,
+            acceptance_criteria=", ".join(task.acceptance_criteria)
+            if task.acceptance_criteria
+            else "N/A",
+            code=chunk.content,
+        )
+        raw = str(Agent(model=resolve_text_mode_strands_model(llm))(prompt)).strip()
```
brandonkindred: **[HIGH] standards** — The `_run_llm_review` function iterates over chunks and for each chunk calls the LLM and parses the response (lines 81-93) without any try/except. If the LLM call fails or the response is malformed, the exception will propagate and abort the entire review, discarding issues from all chunks. This is inconsistent with other agent calls in the codebase that use try/except, and reduces reliability.

**Suggested fix:** Wrap the per-chunk LLM call and parsing in a try/except block. Log the error and continue to the next chunk. Example:
```python
try:
    raw = str(Agent(model=resolve_text_mode_strands_model(llm))(prompt)).strip()
    data = parse_review_template(raw)
    for item in data.get("issues") or []:
        ...
except Exception as exc:
    logger.error("LLM review failed for chunk %d/%d: %s", idx, len(chunks), exc)
    # continue to next chunk
```

backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
```diff
@@ -48,29 +49,50 @@ def _run_llm_review(
...
+          reviewed rather than dropped. Blank files contribute nothing.
+        - Small inputs are reviewed in a single call, as before.
+    """
+    blocks = [(path, content) for path, content in files.items() if content and content.strip()]
+    if not blocks:
+        return []
+    # Budget each prompt at the same per-call size the old code truncated to,
+    # but split at function/method boundaries so no construct is severed and no
+    # file tail is dropped.
+    chunks = list(build_review_chunks(blocks, MAX_REVIEW_CODE_CHARS))
```
brandonkindred, line 82: **[MEDIUM] standards** — The call to `build_review_chunks(blocks, MAX_REVIEW_CODE_CHARS)` splits large files into as many chunks as needed, with no upper bound on the total number of chunks or total characters processed. A very large file could generate dozens of LLM calls, leading to high costs and potential rate limiting. The old truncation at least capped the cost. This could be an operational concern.

**Suggested fix:** Consider adding a maximum number of chunks (e.g., `MAX_REVIEW_CHUNKS = 10`) or a total character budget. If the limit is exceeded, log a warning and skip the remaining chunks, or fall back to truncation. This balances thoroughness with cost control.

brandonkindred (commented):
The PR successfully implements function-aware chunking for code review, with a clean boundary detection module and updated coordinator. The core logic is sound and well-tested. However, there are critical import issues that will cause `ModuleNotFoundError` in standard Python deployments, and high-severity missing error handling in the LLM review fallback. These must be fixed before merge. Additionally, cost control for large files and a few minor code quality improvements are recommended.

**Spec compliance:** The code meets all specified requirements: new `code_boundaries` module with `preferred_break_lines`, coordinator updated to use function boundaries, V2 fallback review functions use chunking instead of truncation, and comprehensive tests are included. The out-of-scope items (QA/Security capped input, frontend doc self-review truncation) are correctly left unchanged.

- `backend/agents/software_engineering_team/tests/test_code_boundaries.py` — **[LOW] testing** — The function `preferred_break_lines` accepts a `path` argument that may be an empty string. The tests do not cover the case where `path=""` to verify it correctly falls back to the heuristic (since `os.path.splitext("")[1]` is `""`, not in `_PYTHON_EXTS`). Adding this test would improve coverage and document expected behavior.

**Suggested fix:** Add a test: `def test_empty_path_uses_heuristic():
    assert preferred_break_lines("", "def f(): pass") == frozenset({1})`

backend/agents/software_engineering_team/frontend_code_v2_team/phases/review.py
```diff
@@ -46,29 +47,50 @@ def _run_llm_review(
...
+          reviewed rather than dropped. Blank files contribute nothing.
+        - Small inputs are reviewed in a single call, as before.
+    """
+    blocks = [(path, content) for path, content in files.items() if content and content.strip()]
+    if not blocks:
+        return []
+    # Budget each prompt at the same per-call size the old code truncated to,
+    # but split at function/method boundaries so no construct is severed and no
+    # file tail is dropped.
+    chunks = list(build_review_chunks(blocks, MAX_REVIEW_CODE_CHARS))
```
brandonkindred, line 80: **[MEDIUM] standards** — The call to `build_review_chunks(blocks, MAX_REVIEW_CODE_CHARS)` splits large files into as many chunks as needed, with no upper bound on the total number of chunks or total characters processed. A very large file could generate dozens of LLM calls, leading to high costs and potential rate limiting. This could be an operational concern.

**Suggested fix:** Consider adding a maximum number of chunks (e.g., `MAX_REVIEW_CHUNKS = 10`) or a total character budget. If the limit is exceeded, log a warning and skip the remaining chunks, or fall back to truncation.

backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
```diff
@@ -0,0 +1,125 @@
...
+        - Returns the start line of every top-level function/class in
+          ``content``; for a decorated construct, the earliest decorator line
+          (so a cut lands above the decorators, never between a decorator and
+          its def).
+        - Returns ``frozenset()`` on ``SyntaxError`` or any other parse failure,
+          so the caller can fall back instead of crashing on partial code.
+    """
+    try:
+        tree = ast.parse(content)
+    except Exception:
```
brandonkindred: **[LOW] standards** — The `except Exception:` clause catches all exceptions, including `MemoryError`, `RecursionError`, and potentially `KeyboardInterrupt`. While the function's contract is to never raise, catching `KeyboardInterrupt` could prevent a user from interrupting a long parse. It's safer to catch only the specific exceptions that `ast.parse` is known to raise (`SyntaxError`, `ValueError`, `RecursionError`) and let unexpected errors propagate.

**Suggested fix:** Change `except Exception:` to `except (SyntaxError, ValueError, RecursionError):` or at least re-raise `KeyboardInterrupt` and `SystemExit`. Example:
```python
except (SyntaxError, ValueError, RecursionError):
    return frozenset()
```

backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+    breaks: set[int] = set()
+    for i, line in enumerate(content.splitlines(), start=1):
+        if not line or not line.strip():
+            continue
+        if line[0].isspace():
+            continue
+        # Leading whitespace is already ruled out above, so the line starts at
+        # column 0 — no lstrip needed before checking the skip prefixes.
+        if line.startswith(_HEURISTIC_SKIP_PREFIXES):
```
brandonkindred, line 132: **[LOW] logic** — The skip prefixes tuple `_HEURISTIC_SKIP_PREFIXES` includes `*/` but not `/*`. A line starting with `/*` (block comment start) will be considered a top-level construct boundary, which could split a comment block from preceding code. This is a known limitation, but adding `/*` to the skip list would be a simple improvement with negligible risk.

**Suggested fix:** Add `"/*"` to `_HEURISTIC_SKIP_PREFIXES` (e.g., `_HEURISTIC_SKIP_PREFIXES = ("}", ")", "]", "*/", "/*", "//", "#")`). This prevents block comment starts from being treated as boundaries.

backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
```diff
@@ -0,0 +1,125 @@
...
+        - Pure: no I/O, no mutation of the arguments or module state.
+    """
+    if not content or not content.strip():
+        return frozenset()
+    ext = os.path.splitext(path or "")[1].lower()
+    if ext in _PYTHON_EXTS:
+        breaks = _python_break_lines(content)
+        if breaks:
+            return breaks
+        # An empty Python result means the source did not parse (e.g. a partial
```
brandonkindred: **[LOW] standards** — The comment on line 54 says 'An empty Python result means the source did not parse (e.g. a partial snippet under review)'. However, an empty result can also occur for valid Python that has no top-level functions/classes (e.g., a script with only top-level statements). The docstring implies that empty always means parse failure, which is misleading.

**Suggested fix:** Update the comment to: 'An empty Python result means either the source did not parse or it contains no top-level functions/classes; in either case, fall through to the heuristic.'

backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
brandonkindred: Fixed in edc91da. Note the bare import does resolve in the current setup (the V2 phase tests import and run `_run_llm_review` fine, and the existing `run_review` already imports `code_review_agent.models` the same way), but I've switched to the fully-qualified `software_engineering_team.code_review_agent.coordinator` path and made it a lazy import inside `_run_llm_review` — matching this file's existing convention for `code_review_agent` imports and not assuming the inner package dir is on `sys.path`.

---

backend/agents/software_engineering_team/frontend_code_v2_team/phases/review.py
brandonkindred: Fixed in edc91da — same change as the backend file: lazy, fully-qualified `from software_engineering_team.code_review_agent.coordinator import build_review_chunks` inside `_run_llm_review`.

---

backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
brandonkindred: Fixed in edc91da — each chunk's LLM call + `parse_review_template` is now wrapped in try/except; a failing chunk is logged (`chunk i/N failed`) and skipped so issues from the other chunks survive. Added `test_run_llm_review_skips_failing_chunk_keeps_others` to cover it.

---

backend/agents/software_engineering_team/frontend_code_v2_team/phases/review.py
brandonkindred: Fixed in edc91da — same per-chunk try/except added to the frontend fallback, with `test_fe_run_llm_review_skips_failing_chunk_keeps_others` covering it.

---

backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
brandonkindred: Addressed in edc91da, but deliberately without a hard cap: capping and skipping the remaining chunks would re-introduce the exact tail-truncation this fallback exists to fix. Instead I added a `logger.warning` when a review fans out past `MANY_CHUNKS_WARN_THRESHOLD` (20) chunks, giving cost/rate-limit visibility while still reviewing every chunk. In practice the chunk count is bounded by the size of one task's changed files, and this path only runs when the primary coordinator agent (which does its own context-aware sizing) is unavailable.

---

backend/agents/software_engineering_team/frontend_code_v2_team/phases/review.py
brandonkindred: Same handling as the backend (edc91da): added the `MANY_CHUNKS_WARN_THRESHOLD` warning for cost visibility, but no hard cap — dropping chunks would reintroduce the truncation this change removes.

---

backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
brandonkindred: Keeping `except Exception:` here. `KeyboardInterrupt` and `SystemExit` derive from `BaseException`, not `Exception`, so they already propagate — this clause does not swallow them. Narrowing to `(SyntaxError, ValueError, RecursionError)` would also re-open the exact gap the earlier [HIGH] review flagged (any *other* `ast.parse` failure would crash the coordinator), and the function's documented contract is to never raise on a parse failure. So `except Exception` is the correct breadth: it catches every parse failure while leaving interrupt/exit signals alone.

---

backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
brandonkindred: Fixed in edc91da — added `"/*"` to `_HEURISTIC_SKIP_PREFIXES` so a column-0 block-comment open is no longer treated as a boundary, and extended `test_heuristic_skips_closing_and_comment_lines` to cover it.

---

backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
brandonkindred: Fixed in edc91da — the comment now notes the empty `ast` result can mean either a parse failure or valid code with no top-level functions/classes, both of which fall through to the heuristic. (The valid-no-constructs case is covered by `test_python_no_constructs_falls_back_to_heuristic`.)

---

backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
```diff
@@ -0,0 +1,130 @@
...
+        - Returns the start line of every top-level function/class in
+          ``content``; for a decorated construct, the earliest decorator line
+          (so a cut lands above the decorators, never between a decorator and
+          its def).
+        - Returns ``frozenset()`` on ``SyntaxError`` or any other parse failure,
+          so the caller can fall back instead of crashing on partial code.
+    """
+    try:
+        tree = ast.parse(content)
+    except Exception:
```
brandonkindred: **[LOW] standards** — The `_python_break_lines` function catches a bare `Exception` to handle parse failures, but it silently swallows all exceptions without any logging. This makes it harder to diagnose unexpected errors (e.g., MemoryError, RecursionError) during development or in production, even though the degradation to no boundaries is intentional.

**Suggested fix:** Add a debug-level log before returning the empty set, for example:
```python
except Exception as exc:
    logger.debug("Python break-line detection failed for %r: %s", path, exc)
    return frozenset()
```
(Import or define `logger` appropriately; the module currently has no logger, so a simple `logging.getLogger(__name__)` would suffice.)

backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
```diff
@@ -40,6 +40,10 @@
 # ---------------------------------------------------------------------------
 
 MAX_REVIEW_CODE_CHARS = 60_000  # Generous limit; review all files, not just first 20
+# A file this many chunks deep means an unusually large review; log a warning
+# for cost/rate-limit visibility, but still review every chunk (dropping any
+# would re-introduce the tail truncation this fallback exists to avoid).
+MANY_CHUNKS_WARN_THRESHOLD = 20
 
 
 def _run_llm_review(
```
brandonkindred, line 49: **[INFO] structure** — The `_run_llm_review` function in `backend_code_v2_team/phases/review.py` is nearly identical to the one in `frontend_code_v2_team/phases/review.py`. This duplication could be refactored into a shared utility to reduce maintenance burden, but it is out of scope for this PR.

**Suggested fix:** Consider extracting the common logic into a shared helper (e.g., in `code_review_agent`) in a follow-up refactoring.

backend/agents/software_engineering_team/frontend_code_v2_team/phases/review.py
```diff
@@ -38,6 +38,10 @@
 # ---------------------------------------------------------------------------
 
 MAX_REVIEW_CODE_CHARS = 60_000  # Generous limit; review all files, not just first 20
+# A file this many chunks deep means an unusually large review; log a warning
+# for cost/rate-limit visibility, but still review every chunk (dropping any
+# would re-introduce the tail truncation this fallback exists to avoid).
+MANY_CHUNKS_WARN_THRESHOLD = 20
 
 
 def _run_llm_review(
```
brandonkindred, line 47: **[INFO] structure** — Same duplication note as above; the frontend `_run_llm_review` mirrors the backend version.

**Suggested fix:** Same as above.

backend/agents/software_engineering_team/tests/test_code_boundaries.py
```diff
@@ -0,0 +1,140 @@
...
+from __future__ import annotations
+
+from code_review_agent.code_boundaries import preferred_break_lines
+
+# ---------------------------------------------------------------------------
+# Python (ast) detection
+# ---------------------------------------------------------------------------
+
+
+def test_python_returns_each_top_level_construct_start() -> None:
```
brandonkindred, line 17: **[LOW] standards** — Test functions in this file lack docstrings. Coding standards require functions to have docstrings explaining purpose, usage, and constraints.

**Suggested fix:** Add a brief docstring to each test function, e.g., for test_python_returns_each_top_level_construct_start: """Verify that preferred_break_lines returns the 1-based start lines of every top-level def, class, and async def in a Python file, excluding nested methods."""

backend/agents/software_engineering_team/tests/test_code_boundaries.py
```diff
@@ -0,0 +1,140 @@
...
+    # An empty path has no extension, so detection falls to the heuristic.
+    assert preferred_break_lines("", "def f(): pass") == frozenset({1})
+
+
+# ---------------------------------------------------------------------------
+# Graceful degradation
+# ---------------------------------------------------------------------------
+
+
+def test_minified_single_line_has_no_interior_breaks() -> None:
```
brandonkindred, line 132: **[LOW] naming** — The function name 'test_minified_single_line_has_no_interior_breaks' contains a typo: 'interior' is not a standard English word; it should be 'interior' (meaning internal) or 'internal'.

**Suggested fix:** Rename to 'test_minified_single_line_has_no_interior_breaks' or 'test_minified_single_line_has_no_internal_breaks'.

backend/agents/software_engineering_team/tests/test_code_review_coordinator.py
```diff
@@ -393,6 +393,83 @@ def test_split_never_breaks_a_line_even_when_oversized() -> None:
...
+    for f in range(1, n_funcs + 1):
+        parts.append(f"function fn_{f:03d}() {{")
+        for b in range(body_lines):
+            parts.append(f"  const value_{f:03d}_{b} = compute({b});  " + "z" * 20)
+        parts.append("}")
+        parts.append("")
+    return "\n".join(parts)
+
+
+def test_split_breaks_at_function_boundary_python() -> None:
```
brandonkindred, line 424: **[LOW] standards** — Test functions in this segment (lines 424, 437, 448, 460) lack docstrings. Coding standards require functions to have docstrings explaining purpose and constraints.

**Suggested fix:** Add a docstring to each test function, e.g., for test_split_breaks_at_function_boundary_python: """Verify that split_block_into_segments cuts a Python file at top-level function boundaries, never mid-body, and that content reassembles exactly."""

backend/agents/software_engineering_team/tests/test_v2_fe_review_phase.py
```diff
@@ -76,6 +76,86 @@ def test_fe_run_llm_review(monkeypatch):
     assert len(issues) == 1
 
 
+def test_fe_run_llm_review_chunks_large_file_without_dropping_tail(monkeypatch):
```
brandonkindred, line 79: **[LOW] standards** — Test functions in this file (new tests at lines 79, 117 and existing tests at lines 159, 178, 198, 227, 258, 287, 349) lack docstrings. Coding standards require functions to have docstrings.

**Suggested fix:** Add a docstring to each test function explaining what behavior it verifies. For example, for test_fe_run_llm_review_chunks_large_file_without_dropping_tail: """Ensure the frontend fallback review splits a file larger than MAX_REVIEW_CODE_CHARS into function-aware chunks so the tail is reviewed instead of truncated."""

backend/agents/software_engineering_team/tests/test_v2_review_phase.py
```diff
@@ -59,6 +59,87 @@ def test_run_llm_review_parses_issues(monkeypatch):
...
```
brandonkindred, line 62: **[LOW] standards** — Test functions in this file (new tests at lines 62, 100 and existing tests at lines 143, 195, 216, 245, 268, 297, 356, 382, 418, 461, 485) lack docstrings. Coding standards require functions to have docstrings.

**Suggested fix:** Add a docstring to each test function. For test_run_llm_review_chunks_large_file_without_dropping_tail: """Verify that the backend fallback review chunks a large Python file at function boundaries so the tail is reviewed, not dropped."""

brandonkindred (commented):
The pull request introduces function-aware chunking to ensure large files are split at construct boundaries rather than mid-body during code review fallback. The new `code_boundaries` module is pure, well-documented, and degrades gracefully. The coordinator's splitter now prefers construct boundaries and falls back to line boundaries when needed, preserving exact reassembly and line numbering. The V2 review phases use the new chunking instead of silent truncation, so the tail of large files is reviewed. The test suite thoroughly covers boundary detection, coordinator splitting, and V2 integration, including edge cases.

Overall, the code is of high quality and meets all acceptance criteria. The review identified only minor issues:

- **Low severity**: In `code_boundaries.py`, the `_python_break_lines` function catches a bare `Exception` to handle parse failures but silently swallows all exceptions without logging, which could hinder diagnosis of unexpected errors. Adding a debug-level log is suggested.
- **Low severity**: Several test functions across `test_code_boundaries.py`, `test_code_review_coordinator.py`, `test_v2_fe_review_phase.py`, and `test_v2_review_phase.py` lack docstrings, which are required by coding standards.
- **Low severity**: A test function in `test_code_boundaries.py` has a typo in its name (`interior` instead of `interior` or `internal`).
- **Informational**: The `_run_llm_review` function is nearly identical in both `backend_code_v2_team/phases/review.py` and `frontend_code_v2_team/phases/review.py`. This duplication could be refactored into a shared utility in a follow-up to reduce maintenance burden.

These findings are consistent with the approved verdict and do not detract from the submission's correctness or completeness.

**Spec compliance:** The submission fully meets the specification and acceptance criteria. Function-aware splitting is implemented, V2 fallback reviews every chunk without dropping the tail, QA/security agents still receive capped strings, and the coordinator preserves exact reassembly and line numbering. The out-of-scope items (#876, #877) are not addressed, as specified. The test suite validates all required behaviors, including boundary detection for Python and other languages, graceful degradation, and integration with the review phases.

backend/agents/software_engineering_team/code_review_agent/code_boundaries.py
brandonkindred: Fixed in e141fdb — added a module logger and a `logger.debug("ast parse failed during break-line detection: %s", exc)` before the `return frozenset()`, so unexpected parse failures stay diagnosable while still degrading to no boundaries.

---

backend/agents/software_engineering_team/backend_code_v2_team/phases/review.py
brandonkindred: Agreed — already tracked as a follow-up in #879 (extract a shared `run_llm_review` parameterized by each team's prompt/parser/`ReviewIssue` factory). Keeping it out of this PR's scope as you note. Resolving.

---

backend/agents/software_engineering_team/frontend_code_v2_team/phases/review.py
brandonkindred: Same as the backend note — tracked in #879 for a follow-up extraction. Resolving.

---

backend/agents/software_engineering_team/tests/test_code_boundaries.py
brandonkindred: Fixed in e141fdb — added a one-line docstring to every test function in `test_code_boundaries.py`.

---

backend/agents/software_engineering_team/tests/test_code_boundaries.py
brandonkindred: No change here — "interior" is a standard English word (adjective: "inner / internal"), and "no interior breaks" is used correctly: the test asserts a one-line minified file has no break points in the *interior* of its content (only line 1, which the splitter never cuts before). The suggested rename targets ("interior" → "interior") are identical, so there's nothing to fix. Resolving.

---

backend/agents/software_engineering_team/tests/test_code_review_coordinator.py
brandonkindred: Fixed in e141fdb — added docstrings to the four function-aware splitter tests this PR added (`test_split_breaks_at_function_boundary_python`/`_typescript`, `_falls_back_to_line_boundary_when_no_constructs`, `_keeps_contiguous_line_bookkeeping`). The other tests in the file are pre-existing and out of scope for this PR.

---

backend/agents/software_engineering_team/tests/test_v2_fe_review_phase.py
brandonkindred: The two tests this PR added here (`test_fe_run_llm_review_chunks_large_file_without_dropping_tail` at line 79 and `test_fe_run_llm_review_skips_failing_chunk_keeps_others`) already carry docstrings. The other tests cited (lines 159, 178, …) are pre-existing and were not touched by this PR, so retrofitting them is out of scope here. Resolving.

---

backend/agents/software_engineering_team/tests/test_v2_review_phase.py
brandonkindred: The two tests this PR added here (`test_run_llm_review_chunks_large_file_without_dropping_tail` at line 62 and `test_run_llm_review_skips_failing_chunk_keeps_others` at line 100) already have docstrings. The remaining cited lines are pre-existing tests not touched by this PR, so they're out of scope. Resolving.

---

